PR #88, opened 2016-09-27
Add regression test and fix 500 error

### LINKED ISSUE #68: /query ohne Parameter abfangen
Wird /query direkt aufgerufen, erscheint ein 500er. Dies sollte abgefangen werden.

diegosteiner:
Fixes #68 

## Steps taken
* Added tests for actions that throw an exception
* Added preconditions to the controller methods that caused the exception

## Result
* When calling without the q param, no error will be thrown. Instead, an empty array will be returned

Note: This PR was made during [renuo open source day 2016](https://www.renuo.ch).

### DIFF
--- a/app/controllers/full_text_controller.rb
+++ b/app/controllers/full_text_controller.rb
@@ -32,6 +32,7 @@ class FullTextController < ApplicationController
   private
 
   def list_people
+    return [] unless params[:q].present?
     entries = Person.search(Riddle::Query.escape(params[:q]),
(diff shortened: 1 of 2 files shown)


RolandStuder:
We will have a look at it probably in the next few weeks, when we have other works on hitobito core. Awesome to see these contributions from your renuo open source day!

RolandStuder:
@diegosteiner O hie, merci viu mau für die Biitrag!

PR #135, opened 2016-12-03
Add missing permissions (group_full and group_and_below_full)

MartinGantenbein:
Note ability should also be permitted to users with group permissions, not only with layer permissions.

### DIFF
--- a/app/abilities/person/note_ability.rb
+++ b/app/abilities/person/note_ability.rb
@@ -17,6 +17,14 @@ class Person::NoteAbility < AbilityDsl::Base
     permission(:layer_and_below_full).
       may(:create, :show).
       in_same_layer_or_below
+
+    permission(:group_full).
+      may(:create, :show).
+      in_same_group
+
+    permission(:group_and_below_full).
+      may(:create, :show).
+      in_same_group_or_below
   end
 
   private
--- a/app/abilities/person_ability.rb
+++ b/app/abilities/person_ability.rb
@@ -22,7 +22,7 @@ class PersonAbility < AbilityDsl::Base
 
     permission(:group_full).may(:show_full, :history).in_same_group
     permission(:group_full).
-      may(:update, :primary_group, :send_password_instructions, :log).
+      may(:update, :primary_group, :send_password_instructions, :log, :index_tags, :manage_tags, :index_notes).
       non_restricted_in_same_group
     permission(:group_full).may(:update_email).if_permissions_in_all_capable_groups
     permission(:group_full).may(:create).all # restrictions are on Roles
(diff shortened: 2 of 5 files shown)


RolandStuder:
Thanks a lot for the pull request!

But we are somewhat reluctant to integrate this. We restricted it to the layer permissions for a reason. Unless we have something as requested in #136 we want to keep this feature available only available to people with administrative functions (that usually have a :layer_full permission).

We would have to clear this change with the current organizations. And this is not really a full solution to the problem of access to notes and tags.

I hope you understand.

We would probably be more willing to integrate something, where this ability can be given to a role independent of the other permissions. If you could add an optional permission :crm (or something like that) that could be added to the role. We could control it in the wagon and would not widen the permission for all organizations.

If you would be willing to do this, please contact me.

ps: @MartinGantenbein Sorry I commented the commit first instead of the pull request.

MartinGantenbein:
I don't fully understand but believe your reasons for restricting to layer permissions are valid (and that's just fine). But I think I would be better off adapting my organization structure to better meet the standard instead of adding an optional permission (I'm afraid there might be similar cases in the future and we'd have to implement optional permissions again and again - which would only add unnecessary complexity).

To fix my structure I'd be happy for some support. We have a quite simple structure with only 3 group types (2 essentially plus the root group):

1. Root: layer = true, children = Club, Team
2. Club: layer = true, childern = Team
3. Team: layer = false, children = Team

Club admins should have full access to the club and all its teams (including sub*teams), team admins should have full access to the team and all sub*teams (but not their parent group). So my admins currently have `:group_and_below_full` permissions and I can't just simply promote them to `:layer_full` (because teams are in the same layer as their parent group and therefore a team admin would gain access to her parent group).
I'd also have to change _layer_ to `true` for teams to get this fixed with layer instead of group permissions. That's fine.

But (and here I need some support): How to handle already existing teams (they stay in the layer they've been created in)?

- Should I manually update the `layer_group_id` field in the `groups` table?
- Any restrictions on `layer_group_id`, how to choose new IDs?
- Or would you suggest creating new teams after setting `layer = true` and then merge the old teams into the new ones?
- Any other ideas on how to move teams into their own layers?

Thanks for your support (and don't hesitate to reject the pull request).

RolandStuder:
@codez Can you give a quick suggestion how to do best convert existing groups to layers?

RolandStuder:
@codez But do not merge this one!

codez:
I think you could simply set `layer = true` for the Teams and then change the corresponding `layer_group_id` to the Team's `id`.

Another solution would be to add/overwrite the abilities in your wagon. This basically means putting your changes into a Module and including it in your `wagon.rb`. Some of the existing wagons already do that. Then your Teams may remain regular groups, i.e. not layers.

@RolandStuder New permissions should be added very reluctantly, a :crm Permission would not be the way to go.

RolandStuder:
@MartinGantenbein Sorry, that we have no nicer solution for this at the moment. But I would really appreciate your input on issue #136. I think we need to do something into this direction. Then we can widen the permissions. I will close the issue for now.

@MartinGantenbein I put you on the invite for our community meeting, that we try to do every 6 months. So don't be surprised if you get an email from us. 😄

PR #141, opened 2016-12-09
changed minute select to only use 5 minute steps

RolandStuder:
Time selects allow to select every minute. However, such precision is not necessary, to have 5, 10, 15 ... available is enough. Even 15min might be enough, as currently we only use times for event dates.

Should I write a test for that?

If yes, where should I put it?


RolandStuder:
This misses a graceful way to handle existing data. Submitting a form with a non matching minute number will set the then number to zero, altering the set time, without informing the user.

PR #152, opened 2016-12-30
Use delegated event handling with turbolinks

MartinGantenbein:
After turbolink navigation, event handlers are missing on the `.person-tag-add` `span`. Using delegated event handling does the trick.


codez:
Thank you very much for the fix!

PR #161, opened 2017-01-12
Fix broken rails console in production environment

severinraez:
Before:

```
RAILS_HOST_NAME=example.com RAILS_ENV=production bundle exec rails c
/home/sraez/.rvm/gems/ruby-2.0.0-p643@pitc-post-innoweek/gems/pry-doc-0.8.0/lib/pry-doc.rb:33:in `<class:Pry>': undefined method `config' for Pry:Class (NoMethodError)
```

After:

```
RAILS_HOST_NAME=example.com RAILS_ENV=production bundle exec rails c
Loading production environment (Rails 4.2.7.1)
2.0.0-p643 :001 > 
```

as well as

```
RAILS_ENV=development bundle exec rails c
require rails:   0.470000   0.120000   0.590000 (  0.580558)
require gems:   2.530000   0.260000   2.790000 (  2.818350)
Loading development environment (Rails 4.2.7.1)
```

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1,8 +1,3 @@
-PATH
-  remote: ../hitobito_insieme
-  specs:
-    hitobito_insieme (0.0.1)
-
 GEM
   remote: https://rubygems.org/
   specs:
@@ -544,7 +539,6 @@ DEPENDENCIES
   haml
   headless
   hirb
-  hitobito_insieme!
   http_accept_language
   jquery-rails
   jquery-turbolinks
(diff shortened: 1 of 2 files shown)


Gemfile.lock
```diff
@@ -544,7 +539,6 @@ DEPENDENCIES
   haml
   headless
   hirb
-  hitobito_insieme!
```
kronn: :+1:

severinraez:
@kronn should be ok now

PR #202, opened 2017-03-18
Display tel: Link with phone numbers

adriananderegg:
enable direct call from phone numbers
hyperlink "tel:012345678" is added to every phone number when displaying a person

this is especially helpfull on mobile devices. 
see also iPhoneURLScheme from apple
https://developer.apple.com/library/content/featuredarticles/iPhoneURLScheme_Reference/PhoneLinks/PhoneLinks.html

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,5 @@
 language: ruby
+cache: bundler
 branches:
   only:
     - master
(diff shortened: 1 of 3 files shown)


RolandStuder:
@kronn Bitte gleich mergen, wenn das ok ist.

adriananderegg:
@kronn spec/decorators/contactable_decorator_spec.rb wie gewünscht angepasst an das neue Format

kronn:
@adriananderegg merci

PR #203, opened 2017-03-18
Improved Person history

lukasblunschi:
- Person history now displays the roles with layers in the group colmmn. This allows to identify to which parent group a group belongs.
- Person history now orders the roles by group with layers instead of only the group name.


RolandStuder:
If good then merge

PR #416, opened 2017-11-22
Reduce warnings of sphinx indexing

kronn:
Locally, I get the following warning:

```
WARNING: key 'charset_type' was permanently removed from Sphinx configuration. Refer to documentation for details.
```

multiple times per index-run. With the removals of this PR, it still works locally, but without the warning.

I cannot determine the reason for this configuration, but since it was the default anyway, I see no harm in removing it.

It could be that there is some hidden reason to keep it. If that is the case, I would like to document this.


kronn:
hmm, the build-failure seems like something I want to solve before merging this. It might be just an undeterministic failure, but I want to make sure. Especially since this is a sphinx-spec.

```
SearchStrategies::Sphinx#list_people as admin finds deleted people
     Failure/Error: expect(result).to include(@deleted_leader)

spec/domain/search_strategies/sphinx_spec.rb:93
```

kronn:
I could not reproduce the aforementioned error. The version we use in production matches the one on my developer machine (2.2).

PR #519, opened 2018-04-06
Fix problem by assuming 00:00 as date_only

diegosteiner:
See #518 for details

### DIFF
--- a/app/domain/export/ics/events.rb
+++ b/app/domain/export/ics/events.rb
@@ -10,22 +10,34 @@ module Export::Ics
 
     def generate(events)
       ical = Icalendar::Calendar.new
-      ical_events = events.map { |event| generate_ical_events(event) }.flatten
+      ical_events = events.map { |event| generate_ical_from_event_dates(event) }.flatten
       ical_events.each { |event| ical.add_event(event) }
       ical.to_ical
     end
 
-    def generate_ical_events(event)
+    def generate_ical_from_event_dates(event)
       event.dates.map do |event_date|
-        Icalendar::Event.new.tap do |ical_event|
-          ical_event.dtstart = event_date.start_at
-          ical_event.dtend   = event_date.finish_at
-          ical_event.summary = "#{event.name}: #{event_date.label}"
-          ical_event.location = event_date.location || event.location
-          ical_event.description = event.description
-          ical_event.contact = event.contact && event.contact.to_s
-        end
+        generate_ical_from_event_date(event_date, event)
       end
+    end
+
+    def generate_ical_from_event_date(event_date, event)
+      Icalendar::Event.new.tap do |ical_event|
+        ical_event.dtstart = datetime_to_ical(event_date.start_at)
+        ical_event.dtend   = datetime_to_ical(event_date.finish_at)
+        ical_event.summary = "#{event.name}: #{event_date.label}"
+        ical_event.location = event_date.location || event.location
+        ical_event.description = event.description
+        ical_event.contact = event.contact && event.contact.to_s
+      end
+    end
+
+    def datetime_to_ical(datetime)
+      return unless datetime.respond_to?(:strftime)
+      if Duration.date_only?(datetime)
+        return Icalendar::Values::Date.new(datetime.strftime(Icalendar::Values::Date::FORMAT))
+      end
+      Icalendar::Values::DateTime.new(datetime.strftime(Icalendar::Values::DateTime::FORMAT))
     end
   end
 end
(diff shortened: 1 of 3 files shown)


app/utils/duration.rb
```diff
@@ -76,7 +80,7 @@ def format_date(value)
   end
 
   def date_only?(value)
```
diegosteiner: Moved this method to make it available outsite of duration, but kept original method in case any of the wagons somehow depend on this (e.g. with ```send```)

app/domain/export/ics/events.rb
```diff
@@ -10,22 +10,34 @@ class Events
...
     end
+
+    def generate_ical_from_event_date(event_date, event)
+      Icalendar::Event.new.tap do |ical_event|
+        ical_event.dtstart = datetime_to_ical(event_date.start_at)
+        ical_event.dtend   = datetime_to_ical(event_date.finish_at)
+        ical_event.summary = "#{event.name}: #{event_date.label}"
+        ical_event.location = event_date.location || event.location
+        ical_event.description = event.description
+        ical_event.contact = event.contact && event.contact.to_s
```
kronn: Maybe I am narrow-minded, but the alignment of the `=` bothers me a little. since this is as non-functional as it gets, I would only change it if you feel the same way.

app/domain/export/ics/events.rb
```diff
@@ -10,22 +10,34 @@ class Events
...
+      end
+    end
+
+    def datetime_to_ical(datetime)
+      return unless datetime.respond_to?(:strftime)
+      if Duration.date_only?(datetime)
+        return Icalendar::Values::Date.new(datetime.strftime(Icalendar::Values::Date::FORMAT))
+      end
```
kronn: have you considered using an `else` instead of returning early?

app/domain/export/ics/events.rb
```diff
@@ -10,22 +10,34 @@ class Events
...
       event.dates.map do |event_date|
-        Icalendar::Event.new.tap do |ical_event|
-          ical_event.dtstart = event_date.start_at
-          ical_event.dtend   = event_date.finish_at
-          ical_event.summary = "#{event.name}: #{event_date.label}"
-          ical_event.location = event_date.location || event.location
-          ical_event.description = event.description
-          ical_event.contact = event.contact && event.contact.to_s
-        end
+        generate_ical_from_event_date(event_date, event)
```
kronn: I like the changes in naming and how the methods have been extracted.

app/utils/duration.rb
kronn: Good thinking. I would defend the removal against anyone, since "thou shalt not `send` for private methods". Then again, someone might have solved a problem this way and this gives an nice upgrade-path. One could add a deprecation warning (https://devdocs.io/rails~4.2/module#method-i-deprecate) to instruct the caller about your new and better way.

spec/domain/export/ics/events_spec.rb
```diff
@@ -32,6 +29,60 @@
...
+      end
+    end
+  end
+
+  describe '#datetime_to_ical' do
+    subject { export.datetime_to_ical(datetime) }
+
+    context 'with fullday event' do
+      let(:datetime) { Time.zone.local(2018, 5, 19) }
+      it { is_expected.to be_a(Icalendar::Values::Date) }
```
kronn: I would use the `let` from above to have the class itself mentioned in fewer places.

spec/domain/export/ics/events_spec.rb
```diff
@@ -32,6 +29,60 @@
...
+    end
+
+    context 'with timed event' do
+      let(:datetime) { Time.zone.local(2018, 5, 19, 12, 15) }
+      it { is_expected.to be_a(Icalendar::Values::DateTime) }
```
kronn: same: I suggest using the `let` above unless some reason stands against it.

app/utils/duration.rb
diegosteiner: 👍

app/domain/export/ics/events.rb
diegosteiner: Alignment was not intentional, getting rid of it.

app/domain/export/ics/events.rb
diegosteiner: Considered it but went for this anyway. I'll change it though.

spec/domain/export/ics/events_spec.rb
diegosteiner: 👍

spec/domain/export/ics/events_spec.rb
diegosteiner: 👍

RolandStuder:
@diegosteiner Big thanks for taking care of this! We really appreciate it!

PR #574, opened 2018-06-20
Tweak wording for event sign up notification

fossil12:
I'm not sure about the impact of this change for other instances, but this is how we would like the success message to look like.

If possible this should be merged until the end of the month an be part of the next release for the Jubla instance.

### DIFF
--- a/config/locales/views.de.yml
+++ b/config/locales/views.de.yml
@@ -425,7 +425,7 @@ de:
   event/participations:
     full_entry_label: '%{model_label} von <i>%{person}</i> in <i>%{event}</i>'
     success: '%{full_entry_label} wurde erfolgreich erstellt. Bitte überprüfe die Kontaktdaten und passe diese gegebenenfalls an.'
-    instructions: 'Für die definitive Anmeldung musst du die Teilnahmebestätigung über <i>Drucken</i> ausdrucken, unterzeichnen und per Post an die entsprechende Adresse schicken.'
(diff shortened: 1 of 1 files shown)


config/locales/views.de.yml
```diff
@@ -425,7 +425,7 @@ de:
   event/participations:
     full_entry_label: '%{model_label} von <i>%{person}</i> in <i>%{event}</i>'
     success: '%{full_entry_label} wurde erfolgreich erstellt. Bitte überprüfe die Kontaktdaten und passe diese gegebenenfalls an.'
```
fossil12: Where does `full_entry_label` come from? We would like to change it to "Die Anmeldung" if it currently says "Teilnahme".

RolandStuder: @fossil12 `full_entry_label` comes from an abstraction layer for a CRUD Controller. See `app/controllers/crud_controller.rb` But as this one is used for multiple models, you should not change it. It's probably best to just create a new decorator method for the participation. See `app/decorators/event/participation_decorator`. You can a new method there `def full_label` which will be available in the views. I also actually think you can remove the part of "bitte überprüfe Kontaktangaben" as not the registration automatically asks for contact details.

RolandStuder:
This will show for PBS and Cevi as well. But you can do this change in the wagon. Just add the correct keys in your organizations wagon locale file file, this will overwrite the text from the core and not affect anyone else. 

So go to `hitobito_jubla/config/locales/views.de.yml` and make the change there, so it will affect no one else.

Everything clear? I think we will not merge this change, as the 14 day deadline might be too specific. But I will immediately merge any of your text changes, if you do it on the jubla wagon.

config/locales/views.de.yml
fossil12: Thank you for details explanation. The impact of this change seems to big to me do any modification right now.

PR #587, opened 2018-06-22
Feature/548 default roles, corrections

RolandStuder:
Was badly implemented, updating a role had automatically selected the default role, instead of the actually already set role, also details on role edit screen was not shown.
also sets default role for imports

### DIFF
--- a/app/controllers/roles_controller.rb
+++ b/app/controllers/roles_controller.rb
@@ -29,6 +29,11 @@ class RolesController < CrudController
 
   hide_action :index, :show
 
+  def new
+    @type = @group.default_role
+    super
+  end
+
   def create
     assign_attributes
     with_person_add_request do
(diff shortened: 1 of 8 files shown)


app/controllers/roles_controller.rb
```diff
@@ -29,6 +29,11 @@ class RolesController < CrudController
 
   hide_action :index, :show
 
+  def new
+    @type = @group.default_role
+    super
```
RolandStuder: Do I have to call super?
Do I need to add arguments? If yes, please correct it and let me know.

app/views/roles/_details.js.haml
```diff
@@ -0,0 +1,3 @@
+- if @type
+  $('#role_info').html('#{escape_javascript(render('info'))}');
+  = render_extensions :details
```
RolandStuder: Had to make this into a partial, as details should be loaded whenever group or role is changed.

PR #619, opened 2018-08-27
Update gems

kronn:
tl;dr There is one security issue and some build-issues. :scream: But this PR fixes them both :smile: 

On our other build-system, we updated the older firefox to a more recent version.
Therefore I took in upon myself to update selenium-webdriver and capybara. 
To reflect and motivate the change for the Community, the `.travis.yml` has been updated as well.

While at the topic of updating Gems, I updated Sprockets to mitigate a security issue and ordered the dependencies in the Gemfile alphabetically to make rubocop more happy.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 language: ruby
 cache: bundler
 addons:
-  firefox: 45.0
+  firefox: 60.0
 branches:
   only:
     - master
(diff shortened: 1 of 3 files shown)


RolandStuder:
@kronn Thanks for the warning about the Ruby Version, I updated the readme and added an entry to the Changelog, so others are not surprised when it should break because of the ruby version.

PR #639, opened 2018-09-29
fix the default role when assigning a new group

Tschet1:
Fix a typo to enable the automatic selection of the default role of a group when assigning a new role to a user.


RolandStuder:
Ooops, thx for that, I will also add some tests :-)

PR #641, opened 2018-09-29
Feature: Display required fields with an asterisk

diegosteiner:
# Problem

As of right now, required fields are indicated with an asterisk _after_ the input-field. This is quiet unconventional and users might oversee it easily.

# Solution

Move the asterisk in front of the input-field, next to the label. Instead of hardcoding it into the label, a "required"-class is added to the control-group. The label is then extended via css.

Credits to @rapkosi


RolandStuder:
Thx, nice!

PR #645, opened 2018-09-30
Unsubscribe link

adriananderegg:
#92 
adds link to unsubscribe page 
differentiates between html/multipart email and plaintext


RolandStuder:
Cool thx, but I guess this needs more work and testing:

* [ ] Currently a person would still need to login to unsubscribe (which of course is still much better than not giving any option). But of course adding the possibility to unsubscribe would need some token per person, which means more stuff we need to save... more complexity.
* [ ] Maybe the html link should have some minimal styles, to make sure it is actually visible (background-color and color and text-decoration).
* [ ] Must be tested with real mails on the integration environment

RolandStuder:
Cant a multipart email also have a plain text part as an alternative to the html part? this should be changed as well? And can't there be a multipart message without html, just a plain text message with an attachement, for example?

adriananderegg:
I'll check on the multipart emails. currently there are no tests defined with multipart emails, so I'll have a look at this in general. 

and include some styles for the link. 

unsubscribe without login: I would there wait for tokenized logins as general feature. also requested in #606

psunix:
wird vom BBT geprüft

psunix:
wir haben die Änderungen in einen neuen PR übernommen: #1160

PR #646, opened 2018-09-30
[WIP] Add contact details to ics export

Tschet1:
**This needs work**

First part of a better description for ics events.
Added information about the creator of the event and added his (public) phone numbers and his standard mail address.
The missing piece is an URL that points directly to the event. The problem here is that the routing component only works for relative url because the _default_url_options[:host]_ is not set. However, I was not able to set this setting. In the end, tests were succeeding but the development environment was not working.
Maybe someone more familiar with the application does know an easy fix, otherwise I will remove the URL and recreate the PR with the contact data only.

### DIFF
--- a/app/domain/export/ics/events.rb
+++ b/app/domain/export/ics/events.rb
@@ -21,13 +21,31 @@ module Export::Ics
       end
     end
 
+    def generate_additional_description_from_event(event)
+      con = event.contact
+      if con
+        con_phone_mail = "\n#{con.person_name}"
+        con.phone_numbers.each do |item|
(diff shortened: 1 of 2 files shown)


diegosteiner:
I'll have look.

RolandStuder:
I assume, that you should be able to get the host from the environment variable 'ENV['RAILS_HOST_NAME']', probably best to set this save this in a config variable like it is done for the mailer in https://github.com/hitobito/hitobito/blob/d615ffff9bde1cc390604b2f8a3eb2a1c8f7499f/config/environments/production.rb#L81

PR #704, opened 2019-01-15
Do not assume multiple parallel requests in development

kronn:
For production and almost anywhere else, it might make sense to tag each
request so that they can be grouped. For development, where I manually
analyse one manual request at a time, I prefer to have less noise.

Also, the log now reads better on a small terminal-window.

Since I might overlook something important here, I put this in a pull-request
with the humble request for comments.


codez:
If you will switch to Puma in development at some point, you might have parallel Ajax Requests. As the majority of requests are still sequential, this might be an insignificant issue.

PR #748, opened 2019-03-31
Fix label-export: Render only company's company_name

sihu:
Label export showed company names even if address was not marked as a company. This is annoying because users often think, they have to fill the `company_name` with the company they currently work. 

If filled like that:

2019-03-31 at 13 17

**Idea:** A further improvement could be, to only display the `company_name` field, if `company` is marked.

**Before:**
2019-03-31 at 13 19

**Now:**
image

### DIFF
--- a/app/domain/export/pdf/labels.rb
+++ b/app/domain/export/pdf/labels.rb
@@ -95,7 +95,7 @@ module Export::Pdf
     end
 
     def print_company?(contactable)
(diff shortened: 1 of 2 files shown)


KunoKunz:
Danke für den Vorschlag. Ich frage noch bei der bestehenden Community, ob es Einwände gibt. Liebi Grüess, André

amaierhofer:
Thanks!

PR #802, opened 2019-07-17
Allow mailcatcher to be defined through ENV

cimnine:
I've setup my local dev-environment using Docker, and my mailcatcher runs in a separate container.
That's why I need a way to overwrite the default configuration for mailcatcher.

### DIFF
--- a/config/environments/development.rb
+++ b/config/environments/development.rb
@@ -40,7 +40,10 @@ Hitobito::Application.configure do
     end
   else
     # mailcatcher
-    config.action_mailer.smtp_settings = { :address => "localhost", :port => 1025 }
+    config.action_mailer.smtp_settings = {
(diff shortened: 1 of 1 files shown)


kronn:
I think this could also be achieved by setting existing variables accordingly:

```
RAILS_MAIL_DELIVERY_METHOD=smtp
RAILS_MAIL_DELIVERY_CONFIG="{address: 'localhost', port: 1025}"
```

Would that be working for your use-case?

cimnine:
It works, when I set the values as follows:

```
RAILS_MAIL_DELIVERY_METHOD=smtp
RAILS_MAIL_DELIVERY_CONFIG=address: mail, port: 1025
```

`mail` is the hostname of the mailserver/mailcontainer.

PR #823, opened 2019-09-07
Update picture_uploader.rb

diegosteiner:
Profile pictures have limited use when they're too small. 

- [Twitter recommends profile pictures that are > 400px: rake db:migrate wagon:migrate](https://help.twitter.com/en/managing-your-account/common-issues-when-uploading-profile-photo)


kronn:
Looks good to me now. Thanks for this.

PR #825, opened 2019-09-08
Make first column of people view have flexible with.

cleverer:
Unfortunately Safari rendering engine rounds differently than others, so the second column of the people view always got wrapped on to a new line. By making the first on grow to use available space, the problem won't occur anymore.

Tested in current versions of Safari, Firefox, Chrome, Opera, IE 11 on macOS, Windows and Linux.


carlobeltrame:
Thank you! Finally done.

PR #838, opened 2019-09-21
Change icon for accepting request

lukaseisenring:
Change the icon for accepting an request because the current icon does not exist. Makes the function working again.

Current state:
image

After PR:
image


amaierhofer:
Thanks for fixing!

PR #839, opened 2019-09-22
Fix search order for group logo.

cleverer:
We had an example as follows:

Pfadi Züri (Kantonalverband) added a Logo. I added a Logo to Pfadi Neuburg/Wartensee (Abteilung subordinate to Pfadi Züri). My Logo never showed up (instead showing Pfadi Züri), as the search began in the top of the tree…
Now it works as expected.


kronn:
Thanks for the bugfix (well: my bug, your fix :wink: ).

I would like to have a spec for this to prevent unintentionally breaking this again in the future. Could you write such a spec?

cleverer:
I'm not very experienced with Ruby yet, but I can try to do that when I find time. But I can't promise to have anything anytime soon.

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

cleverer:
I started working on such a spec, but as I feared the project doesn't seem to be novice-friendly enough for me.

Right now I'm struggling on loading a model (or creating a mock) to test the desired behaviour. If someone could give me some tips, I might be able to finish this.

In the meantime, I'm going to try to introduce myself to rails/rspec some more…

PR #841, opened 2019-09-22
Add cypress-on-rails from hitobito_pbs wagon

cleverer:
Required by: https://github.com/hitobito/hitobito_pbs/pull/110

Adds the cypress-on-rails dependency to lockfile.


carlobeltrame:
Wagon dependencies can be added in the wagon's gemspec file without modifying the core. An example can be seen here: https://github.com/hitobito/hitobito_sbv/blob/master/hitobito_sbv.gemspec#L20

cleverer:
> Wagon dependencies can be added in the wagon's gemspec file without modifying the core. An example can be seen here: https://github.com/hitobito/hitobito_sbv/blob/master/hitobito_sbv.gemspec#L20

That's actually the thing I did. Except that I didn't specify a version…

PR #871, opened 2019-11-09
Style consent screen to match layout

diegosteiner:
image

See https://github.com/doorkeeper-gem/doorkeeper/wiki/Authorization-Code-Flow


amaierhofer:
Thanks!

PR #935, opened 2020-03-18
Use global CSRF token when changing action in participation_lists form

simfeld:
fixes hitobito/hitobito#934


carlobeltrame:
Thanks a lot!

PR #982, opened 2020-07-07
Move Log-Tagging to production only

kronn:
Essentially: move it away from development, because there is it not as
helpful. Also, it generates a lot of noise in the test-log that no-one
ever will look at.

Am I missing something here?


amaierhofer:
@carlobeltrame  genau, Integration ist nur die Stage, aber dort läuft die app ebenfalls im environment production

PR #994, opened 2020-08-12
Set root password in development

carlobeltrame:
Replaces #993 
Better to only do this in the development seeds. Thanks @kronn 
The other fields (firstname, lastname, birth date) are necessary to set because they are required in some wagons.

See hitobito/development#3

### DIFF
--- a/db/seeds/development/root.rb
+++ b/db/seeds/development/root.rb
@@ -1,0 +1,8 @@
+# encoding: utf-8
+
+#  Copyright (c) 2020, Puzzle ITC. This file is part of
+#  hitobito and licensed under the Affero General Public License version 3
+#  or later. See the COPYING file at the top-level directory or at
+#  https://github.com/hitobito/hitobito.
+
+Person.first.update!(password: 'hito42bito', first_name: 'Puzzle', last_name: 'ITC', birthday: '1999-09-09')


db/seeds/development/root.rb
```diff
@@ -0,0 +1,8 @@
+# encoding: utf-8
+
+#  Copyright (c) 2020, Puzzle ITC. This file is part of
+#  hitobito and licensed under the Affero General Public License version 3
+#  or later. See the COPYING file at the top-level directory or at
+#  https://github.com/hitobito/hitobito.
+
+Person.first.update!(password: 'hito42bito', first_name: 'Puzzle', last_name: 'ITC', birthday: '1999-09-09')
```
kronn, line 8: Why not use 1970-01-01 as obvious "not null NULL"-value? Then again 1999-09-09 ist also a nice hint and maybe avoids alienation with a younger audience, while being old enough to be considered a grown up in western europe.

carlobeltrame: As far as I can tell (it's not quite clear in the sources I had), 1999-09-09 is the founding date of Puzzle ITC. That's why I chose it.

kronn: Makes sense and I totally missed that.

PR #998, opened 2020-08-18
Human-readable VERSION for Sentry

kronn:
While the OPENSHIFT_BUILD_COMMIT is technically correct and unique, it is far from human-parseable. Since we now update the VERSION with every release, we can use that as release-information.

### DIFF
--- a/config/application.rb
+++ b/config/application.rb
@@ -111,7 +111,9 @@ module Hitobito
     end
 
     def self.versions(file = Rails.root.join('WAGON_VERSIONS'))
-      @versions ||= file.exist? ? file.read.lines.reject(&:blank?) : []
+      @versions ||= {}
+      @versions[file] ||= (file.exist? ? file.read.lines.reject(&:blank?) : nil)
+      @versions[file].to_a
     end
 
     def self.build_info
--- a/config/initializers/sentry.rb
+++ b/config/initializers/sentry.rb
@@ -1,9 +1,7 @@
 Raven.configure do |config|
   config.sanitize_fields = Rails.application.config.filter_parameters.map(&:to_s)
 
-  if ENV['OPENSHIFT_BUILD_COMMIT']
-    config.release = ENV['OPENSHIFT_BUILD_COMMIT']
-  end
+  config.release = Rails.application.class.versions(Rails.root.join('VERSION')).first.chomp
 
   if ENV['SENTRY_CURRENT_ENV'].blank?
     project, stage = [


kronn:
I am looking forward to arguments against this. Something I missed, some better way of doing this.

config/application.rb
```diff
@@ -111,7 +111,9 @@ def self.sphinx_local?
     end
 
     def self.versions(file = Rails.root.join('WAGON_VERSIONS'))
-      @versions ||= file.exist? ? file.read.lines.reject(&:blank?) : []
+      @versions ||= {}
```
amaierhofer, line 114: The only piece of code I found accessing `Rails.application.class.version` is in the sentry initializer. So the default argument `WAGON_VERSIONS` will not be read at any time, therefor could be removed

config/initializers/sentry.rb
```diff
@@ -1,9 +1,7 @@
 Raven.configure do |config|
   config.sanitize_fields = Rails.application.config.filter_parameters.map(&:to_s)
 
-  if ENV['OPENSHIFT_BUILD_COMMIT']
-    config.release = ENV['OPENSHIFT_BUILD_COMMIT']
-  end
+  config.release = Rails.application.class.versions(Rails.root.join('VERSION')).first.chomp
```
amaierhofer, line 4: Would it make sense to keep using the BUILD on integration and only read VERSION for production releases?

amaierhofer (commented):
Improving this would be good, not sure about `WAGONS_FILE` or if sentry is capable of dealing with multiple lines for `config.release`

I think having the contents of `VERSION` for production deployments would make sense, for integration the build number would make more sense

config/initializers/sentry.rb
kronn: IMHO it would not make sense. Theoretically, one could go an find the exact build with this and try to reproduce this locally. I have never done this and cannot imagine doing it. The version however gives me a useful approximation to decide how fixed an older bug might be and where to check in the source.

For production, every build should have a different tag. So the build-commit is mostly the same, just less readable.

kronn: If you find the build-commit  useful as additional tag, we can add that back in. PullRequests are welcome :-)

config/application.rb
kronn: I thought I would be used in the footer where we display the deployed wagon-versions and link the respective commits. If that is wrong, then it could be indeed removed.

PR #1010, opened 2020-08-30
OAuth authorizations German spelling

simfeld:
The correct German spelling for authorizations is 'Autorisierungen' (without the 'h', I'm also confused by this). These are the only two places where I found it misspelt. Compare https://www.duden.de/rechtschreibung/Autorisierung.


carlobeltrame (approved):
Thanks!

PR #1036, opened 2020-09-27
Add specs for group_list_serializer

diegosteiner:
Add the group type to the groups list. 

image

There might be reasons for this not to be desirable, so it needs to be discussed and thought through.


amaierhofer:
Thanks

PR #1042, opened 2020-09-29
Implement e-mail validator for people, refs hitobito/hitobito_cvp#50

psunix:


app/domain/contactable/email_validator.rb
```diff
@@ -0,0 +1,50 @@
...
+          tag_invalid!(p, p.email)
+        end
+        validate_additional_emails(p)
+      end
+    end
+
+    private
+
+    def validate_additional_emails(person)
+      additional_emails = person.additional_emails.pluck(:email)
```
carlobeltrame, line 18: Should we extract `person.additional_emails.pluck(:email)` into a separate method, so that wagons can more easily customize which of their email fields are validated? And if so, could we somehow generalize this so `p.email` and additional emails are in the same list and we don't have to treat them separately?

app/domain/person_tags/validation.rb
```diff
@@ -0,0 +1,43 @@
...
+        find(EMAIL_PRIMARY_INVALID, create: create)
+      end
+
+      def email_additional_invalid(create: false)
+        find(EMAIL_ADDITIONAL_INVALID, create: create)
+      end
+
+      private
+
+      def find(tag, create: false)
```
carlobeltrame, line 29: The `create` switch completely determines the behaviour of this method, and in the code that uses it, it is always clear which one to use. Consider splitting it into two separate methods `find` and `create_if_not_exists`, both without the `create` argument.

app/helpers/tag_lists_helper.rb
```diff
@@ -15,7 +10,20 @@ def available_tags_checkboxes(tags)
     end, '')
   end
 
-  private
```
carlobeltrame, line 18: Restore `private` in an appropriate place, so that the methods below your new ones don't get exposed

app/helpers/tag_lists_helper.rb
```diff
@@ -15,7 +10,20 @@ def available_tags_checkboxes(tags)
...
+  def format_tag_category(category)
+    case category
+    when :other
+      t('.category_other')
+    when :category_validation
+      t('.category_validation')
+    else
+      category
+    end
```
carlobeltrame, line 21: Consider using a hash for this mapping, like you did in PersonTags::Translator

app/models/concerns/tooltip_for_tags.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+module TooltipForTags
+
+  extend ActiveSupport::Concern
+
+  # tried to add this virtual attr to a decorater, but I was unable
+  # to decorate ActsAsTaggableOn::Tag
```
carlobeltrame, line 8:
```suggestion
  # We cannot decorate ActsAsTaggableOn::Tag, so we have to add the attribute this way
```

app/models/person_tags/validation_tagged.rb
```diff
@@ -0,0 +1,21 @@
...
+
+  private
+
+  # if person can be updated, all validations passed. so we asume it's save to remove
+  # all validation tags which marks this person for having invalid attributes
+  def remove_validation_tags
+    tags = PersonTags::Validation.list
+    if tags.present?
+      ActsAsTaggableOn::Tagging.where(taggable: self, tag: tags).destroy_all
+    end
```
carlobeltrame, line 19: This circumvents the permission system in the sense that a Person normally cannot edit their own tags, but I think it is okay here. Just something to be aware of.

app/views/subscriptions/_subscription.html.haml
```diff
@@ -14,7 +14,7 @@
   - unless subscription.tag_list.empty?
     %p
       = t('.only_persons_with')
-      - subscription.tag_list.each do |tag|
-        %span.label.label-inverse.person-tag= tag
+      - subscription.tags.each do |tag|
```
carlobeltrame, line 17: I don't understand why this change from `tag_list` to `tags` is necessary. Does `tag_list` only contain the names while `tags` returns the `Tag` models? Or what else is the reason?

carlobeltrame (changes requested):

app/models/concerns/categorized_tags.rb
```diff
@@ -35,8 +35,14 @@ def grouped_by_category
         h[category] ||= []
         h[category] << tag
       end
+      order_categorized(tags)
+    end
```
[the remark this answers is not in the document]
psunix: yes, inside person/tags_controller

app/controllers/people_filters_controller.rb
```diff
@@ -82,7 +77,7 @@ def people_list_path(options = {})
   end
 
   def load_possible_tags
-    @possible_tags ||= Person.tags
+    @possible_tags ||= PersonTags::Translator.new.possible_tags
```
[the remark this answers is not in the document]
psunix: can't see any difference in numbers of loaded entries here.
since we're using tags only for people, it should not load more than before.
Am I missing something ?

app/controllers/person/tags_controller.rb
```diff
@@ -47,6 +47,14 @@ def destroy
 
   private
 
+  def person_tags
+    @person
+      .reload
+      .tags
+      .order(:name)
+      .grouped_by_category
```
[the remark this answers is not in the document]
psunix: yes! this one is for displaying at person#show.

app/helpers/tag_lists_helper.rb
psunix: well, methods inside helpers are exposed anyway. it's more informative to have it declared private. Going to restore it

app/helpers/tag_lists_helper.rb
psunix: I don't like having to much methods in helper modules, since all methods are exposed anyway.

app/views/subscriptions/_subscription.html.haml
psunix: yes, that's exactly the difference! we need instances of Tag for the translation helper

PR #1101, opened 2020-12-04
Allow editing group / role subscriptions in mailing lists

carlobeltrame:
So far, subscriptions in mailing lists can only be created and deleted. In the case of group subscriptions, entering the data is tedious, and also prone to errors and change over time, when new groups are founded, roles are newly used in a group, or new roles are even added to hitobito. This PR adds the capability to edit group subscriptions.

Thanks to @tobifra for the suggestion :)


amaierhofer:
Thanks

PR #1124, opened 2021-01-08
Feature/cvp 59 subscriber inverse tag filter

TheWalkingLeek:
Resolves https://github.com/hitobito/hitobito_cvp/issues/59


psunix:
Anstatt: Personen mit einem der untenstehenden Tags ausschliessen: 
-> Auschliessen von Personen mit folgenden Tags:

Anstat: Nur Person mit einem der untenstehenden Tags hinzufügen:
-> Hinzufügen von Personen mit folgenden Tags:

app/domain/mailing_list/subscribers.rb
```diff
@@ -116,4 +109,25 @@ def event_subscribers(condition)
                  Event.sti_name,
                  true)
   end
+
+  private
+
+  def excluded_people_by_tag
```
psunix, line 115: since you're returning a list of ids, it makes sense to include keyword ids in method name.

tag_excluded_person_ids ?

app/domain/mailing_list/subscribers.rb
```diff
@@ -116,4 +109,25 @@ def event_subscribers(condition)
...
+    ActsAsTaggableOn::Tagging.select(:taggable_id)
```
psunix, line 116: in my experience it's more common to use pluck instead of select.
but select is probably equal. what do you think?

app/domain/mailing_list/subscribers.rb
```diff
@@ -116,4 +109,25 @@ def event_subscribers(condition)
...
+  end
+
+  def excluded_subscription_tags
+    SubscriptionTag.select(:tag_id)
+                   .joins(:subscription)
+                   .where(subscription_tags: { excluded: true },
+                                             subscriptions: { mailing_list_id: id })
+  end
+
+  def excluded_person_subscribers
```
psunix, line 127: subscription_ids ?

app/domain/mailing_list/subscribers.rb
```diff
@@ -116,4 +109,25 @@ def event_subscribers(condition)
...
+    Subscription.select(:subscriber_id).
+      where(mailing_list_id: id,
+            excluded: true,
+            subscriber_type: Person.sti_name)
```
psunix, line 131: nice, you're using sti_name method here. please do the same on line 117 :wink:

app/domain/mailing_list/subscribers.rb
```diff
@@ -116,4 +109,25 @@ def event_subscribers(condition)
...
   end
+
+  private
+
+  def excluded_people_by_tag
+    ActsAsTaggableOn::Tagging.select(:taggable_id)
+                             .where(taggable_type: 'Person', tag_id: excluded_subscription_tags)
+  end
+
+  def excluded_subscription_tags
```
psunix, line 120: tag_excluded_subscription_ids ?

app/domain/person/subscriptions.rb
```diff
@@ -21,6 +21,7 @@ def direct
 
   def exclusions
```
psunix, line 22: does it make sense to cache this value since it's now used more than once in this class?

app/models/subscription_tag.rb
```diff
@@ -0,0 +1,23 @@
+
+#  Copyright (c) 2012-2020, CVP Schweiz. This file is part of
+#  hitobito_cvp and licensed under the Affero General Public License version 3
+#  or later. See the COPYING file at the top-level directory or at
+#  https://github.com/hitobito/hitobito_cvp.
```
psunix, line 5: it's core ... not hitobito_cvp

app/views/subscriber/group/_tags.html.haml
```diff
@@ -6,12 +6,25 @@
...
-             :second,
-             :first,
-             { },
-             { class: 'chosen-select',
-               multiple: true,
-               data: { chosen_no_results: t('global.chosen_no_results'),
-                       placeholder: ' ' } })
+  = f.collection_select(:included_subscription_tags,
+                        @possible_tags,
+                        :third,
```
psunix, line 11: I still don't like :third here ...

deluxe version would be to have getters for id and name.

maybe it's to extensive ... but how about creating a TagDecorator with #label method which does the translation ?
we could probably remove the whole Translator class ?

ActsAsTaggableOn::Tag.all.order(:name).decorate ?

I think it's a good idea for you to check this option and get some knowledge about decorators in general.

app/views/subscriptions/_subscription.html.haml
```diff
@@ -11,10 +11,15 @@
       - groups.each do |group, role_types|
         %dt= group
         %dd= role_types.map(&:label).join(', ')
-  - unless subscription.tag_list.empty?
+  - unless subscription.subscription_tags.included.empty?
```
psunix, line 14: I would rather use if ... present? instead of unless ... empty?

config/locales/views.fr.yml
```diff
@@ -1170,6 +1171,7 @@ fr:
       export_not_allowed: 'Tu n''as pas le droit d''exporté les abonnés de cet abonnement, car ils contiennent des rôles ou des activités qui ne font pas partie de ton domaine d''autorisation.'
     subscription:
       only_persons_with: 'Seulement personnes avec:'
+      persons_excluded_with: 'Personnes exclues avec:'
```
psunix, line 1174: usually we only translate the German version and the translation is done by translators. but of course, if it's short text and you know what you're doing, then do it :)

db/migrate/20210108142958_add_subscription_tags.rb
```diff
@@ -0,0 +1,15 @@
+class AddSubscriptionTags < ActiveRecord::Migration[6.0]
+  def up
+    create_table :subscription_tags do |t|
+      t.boolean :excluded
+    end
+
+    add_reference :subscription_tags, :subscription, foreign_key: true, type: :integer
```
psunix, line 7: use t.references in block ?

db/migrate/20210108142958_add_subscription_tags.rb
```diff
@@ -0,0 +1,15 @@
+class AddSubscriptionTags < ActiveRecord::Migration[6.0]
```
psunix, line 1: license header ?

spec/controllers/people_filters_controller_spec.rb
```diff
@@ -44,8 +44,8 @@
         Contactable::EmailValidator.new.validate_people
 
         get :new, params: { group_id: group.id }
-        invalid_email_tags = [['Haupt-E-Mail ungültig', 'category_validation:email_primary_invalid'],
-                              ['Weitere E-Mail ungültig', 'category_validation:email_additional_invalid']]
+        invalid_email_tags = [['Haupt-E-Mail ungültig', 'category_validation:email_primary_invalid', PersonTags::Validation.email_primary_invalid.id],
```
psunix, line 47: :+1:

db/migrate/20210108142958_add_subscription_tags.rb
```diff
@@ -0,0 +1,15 @@
...
+  def up
+    create_table :subscription_tags do |t|
+      t.boolean :excluded
+    end
+
+    add_reference :subscription_tags, :subscription, foreign_key: true, type: :integer
+    add_reference :subscription_tags, :tag, foreign_key: true, type: :integer
```
psunix, line 8: not_null ?

spec/models/mailing_list_spec.rb
```diff
@@ -515,4 +530,12 @@ def create_subscription(subscriber, excluded = false, *role_types)
     sub
   end
 
+  def subscription_tags(names)
+    tags = names.map do |name|
+      ActsAsTaggableOn::Tag.create_or_find_by!(name: name)
+    end
+    tags.map do |tag|
+      SubscriptionTag.create!(tag: tag)
```
psunix, line 538: this creates a SubscriptionTag entry without a subscription_id ... use new instead ?
if you're going to make subscription_id not null, this won't work anyway :wink:

psunix (changes requested):
chunnt guät :dancers: 

some things to check!

app/domain/mailing_list/subscribers.rb
carlobeltrame: I think `.pluck` immediately fetches the records, while `.select` is used to just declare the column selection, but still keep it a pending query. So if you change to `.pluck`, it should go after the `.where`.

TheWalkingLeek: Right. On line 21, I call `to_sql` to this statement. This way, theres only one sql statement being queried in the end, because this statement gets embedded into the main query.
This was already done for the `excluded_person_subscribers` and I figured this was because of performance reasons but I may be wrong about that.

psunix: > I think `.pluck` immediately fetches the records, while `.select` is used to just declare the column selection, but still keep it a pending query. So if you change to `.pluck`, it should go after the `.where`.

of course, thx @carlobeltrame for clarifying :flashlight:

psunix: > Right. On line 21, I call `to_sql` to this statement. This way, theres only one sql statement being queried in the end, because this statement gets embedded into the main query.
> This was already done for the `excluded_person_subscribers` and I figured this was because of performance reasons but I may be wrong about that.

then it makes sense to use `.select` here :fireworks:

app/controllers/subscriber/group_controller.rb
```diff
@@ -82,5 +82,17 @@ def replace_validation_errors
       end
     end
 
+    def subscription_tags
+      [collect_tags, collect_tags(excluded: true)].flatten.map do |tag|
```
psunix, line 86: `[collect_tags, collect_tags(excluded: true)]` -> create method or variable all_tags / tags ? 

tags = collect_tags + collect_excluded_tags

makes it more readable

app/controllers/subscriber/group_controller.rb
```diff
@@ -82,5 +82,17 @@ def replace_validation_errors
...
+        SubscriptionTag.new(subscription: entry,
+                            tag_id: tag[:id],
+                            excluded: tag[:excluded]) unless tag[:id].empty?
+      end.compact
+    end
+
+    def collect_tags(excluded: false)
+      param_key = excluded ? :excluded_subscription_tags : :included_subscription_tags
+      model_params[param_key]&.map { |tag| { id: tag, excluded: false } } || []
```
psunix, line 95: create own method collect_excluded_tags ?

`{ id: tag, excluded: false }` shouldn't this be `{ id: tag, excluded: excluded }` ?

app/domain/person/subscriptions.rb
```diff
@@ -32,19 +33,26 @@ def from_groups
...
       .joins(:related_role_types)
-      .left_joins(:tags)
+      .left_joins(:subscription_tags)
       .where(condition.to_a)
+      .where.not(id: exclusions.pluck(:id))
+  end
+
+  private
+
+  def exclusions_by_subscription_tags
```
psunix, line 55: returns a list of ids ... so include ids in method name :wink:

app/models/subscription.rb
```diff
@@ -84,6 +84,14 @@ def grouped_role_types
     result
   end
 
+  def included_subscription_tags
```
psunix, line 87: iiiiiiiids :dancers:

app/views/subscriber/group/_tags.html.haml
psunix: how about my last comment ?

app/views/subscriptions/_subscription.html.haml
```diff
@@ -11,10 +11,15 @@
       - groups.each do |group, role_types|
         %dt= group
         %dd= role_types.map(&:label).join(', ')
-  - unless subscription.tag_list.empty?
+  - if subscription.subscription_tags.included.any?
```
psunix, line 14: `present?` instead of `any?`

db/migrate/20210114074037_add_subscription_tags.rb
```diff
@@ -0,0 +1,20 @@
+#  Copyright (c) 2012-2020, CVP Schweiz. This file is part of
+#  hitobito_cvp and licensed under the Affero General Public License version 3
+#  or later. See the COPYING file at the top-level directory or at
+#  https://github.com/hitobito/hitobito.
+
+class AddSubscriptionTags < ActiveRecord::Migration[6.0]
+  def up
```
psunix, line 7: `change` instead of up/down ?

db/migrate/20210114074037_add_subscription_tags.rb
```diff
@@ -0,0 +1,20 @@
...
+    create_table :subscription_tags do |t|
+      t.boolean :excluded
```
psunix, line 9: boolean attrs should always have a default value, not null

spec/models/mailing_list_spec.rb
```diff
@@ -515,4 +530,12 @@ def create_subscription(subscriber, excluded = false, *role_types)
     sub
   end
 
+  def subscription_tags(names)
+    tags = names.map do |name|
+      ActsAsTaggableOn::Tag.create_or_find_by!(name: name)
+    end
+    tags.map do |tag|
```
psunix, line 537: on-liner with `{}` ?

psunix (changes requested):
almost there ... go go go :rocket:

Gemfile.lock
```diff
@@ -1,7 +1,7 @@
 PATH
-  remote: ../hitobito_generic
+  remote: ../hitobito_cvp
```
psunix, line 2: bitte immer darauf achten das hier immer der generic wagon eingecheckt wird.

db/migrate/20210114074037_add_subscription_tags.rb
psunix: ?

psunix: ah ... now change is not possible anymore

db/migrate/20210114074037_add_subscription_tags.rb
```diff
@@ -0,0 +1,32 @@
...
+    create_table :subscription_tags do |t|
+      t.boolean :excluded, default: false
+
+      t.references :subscription, foreign_key: true, type: :integer, null: false
+      t.references :tag, foreign_key: true, type: :integer, null: false
+    end
+
+    Subscription.groups.find_each do |subscription|
+      tagging = ActsAsTaggableOn::Tagging.find_by(taggable_id: subscription.id,
+                                                   taggable_type: Subscription.sti_name)
```
psunix, line 17: shouldn't we also remove existing tagging entry after migrating ?

psunix: maybe it's a good idea to write a spec for this migration. 
example: https://github.com/hitobito/hitobito/blob/messages/spec/migrations/migrate_mail_log_spec.rb

app/controllers/subscriber/group_controller.rb
```diff
@@ -82,5 +82,23 @@ def replace_validation_errors
...
 
+    def subscription_tags
+      tags = collect_included_tags + collect_excluded_tags
+      tags.map do |tag|
+        next if tag[:id].empty?
+
+        SubscriptionTag.new(subscription: entry,
+                            tag_id: tag[:id],
+                            excluded: tag[:excluded])
+      end.compact
```
psunix, line 93: compact still required ?

psunix (approved):
looks great ! rebase && merge :dancers:

spec/migrations/add_subscription_tags_spec.rb
```diff
@@ -0,0 +1,82 @@
...
+                                        tag_id: email_additional_invalid.id,
+                                        context: 'tags')
+
+      expect(ActsAsTaggableOn::Tagging.count).to eq(2)
+
+      expect do
+        migration.up
+      end.to change { ActsAsTaggableOn::Tagging.count }.by(-2)
+
+      expect(SubscriptionTag.where(subscription: subscription).count).to eq(2)
```
psunix, line 50: subscription.subscriptions_tags.count ? (maybe reload)

psunix (approved):

PR #1140, opened 2021-01-28
Enable duplicating existing letters

carlobeltrame:
Fixes https://github.com/hitobito/hitobito_cvp/issues/99

This allows to use an existing letter as a template for a new one, even after the existing one has been dispatched and is read-only.

### DIFF
--- a/app/controllers/messages_controller.rb
+++ b/app/controllers/messages_controller.rb
@@ -30,7 +30,21 @@ class MessagesController < CrudController
 
   before_render_form :set_recipient_count
 
+  before_action :authorize_duplicate, only: :new
+
+  def new
+    assign_attributes_from_duplication_source if duplication_source.present?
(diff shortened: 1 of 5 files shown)


app/controllers/messages_controller.rb
```diff
@@ -29,6 +43,10 @@ def type
     model_params && model_params[:type].presence
   end
 
+  def template
```
carlobeltrame, line 46: Rename to `duplication_source`

app/controllers/messages_controller.rb
```diff
@@ -29,6 +43,10 @@ def type
...
+    @template ||= Message.find_by id: params[:duplicate_from]
```
carlobeltrame, line 47: Rename param to `:duplication_source_id`

PR #1161, opened 2021-02-22
Standard rubocop

kronn:
Building upon https://github.com/testdouble/standard and https://evilmartians.com/chronicles/rubocoping-with-legacy-bring-your-ruby-code-up-to-standard I suggest that we take our RuboCop-Usage to a level where it actually helps.

This PR
- introduces standard.rb
- changes the config to integrate standard's rules with rubocop
- _is therefore making every existing rubocop-integration a working standard-integration_
- has a few "auto-correct" commits to start with ... clean-ish slate.
- adds a bin-stub so the right versions of standard and rubocop can be easily used
- disables and removes most of our previous choices
- introduces an `ApplicationRecord`
- does intentionally *NOT* add a `.rubocop-todo.yml`
- moves our strict "must"-rules into `.rubocop-strict.yml`
- **awaits your opinions**
- will result in follow-up PRs for the wagons to keep things in sync there are well, iff you all accept and we decide to proceed with this approach


rnestler:
@kronn I think this would be valuable. Why was it abandoned?

kronn:
No-one wanted to discuss it, then it went out-of-date. I plan to rebase the ideas and recreate this PR this or next week.

rnestler:
As a source for inspiration: At Renuo we created our own cop gem based on `standardrb` and added some custom configuration: https://github.com/renuo/renuocop / https://github.com/renuo/renuocop/blob/main/config/base.yml

kronn:
Sadly, your RenuoCop is not opensource, but I know and like the idea.

kronn:
Done through #2727

Issue #1190, opened 2021-03-06
TECH: Von Travis auf GitHub Actions migrieren


BacLuc:
Follow up to #1088
To do / Tech spec:

- [x] create hitobito Github Action (done #1191)
- [x] create hitobito_youth Github Action (done hitobito/hitobito_youth#14)
- [x] create hitobito_pbs Github Action (done in hitobito/hitobito_pbs#190)
- [x] cache the dependencies in vendor/bundle (done in the above PRs)
- [x] create Github Action for other Wagons
- [x] delete travis.yml from the repos

Possible further improvements

- [ ] Set Github Action checks as required
- [ ] Gemfile.lock should match the Gemfile
- [ ] add dependency to rake to Gemfile
- [ ] Pin version of node and yarn
- [ ] Update batch in readme.md

PR #1191, opened 2021-03-06
Add GitHub Action for CI

BacLuc:
Issue: #1190

für core parallel github actions aktivieren


BacLuc:
@carlobeltrame are Github Actions deactivated on this repo?
In my fork the github actions were executed.

.github/workflows/tests.yml
```diff
@@ -0,0 +1,87 @@
...
+      RAILS_DB_USERNAME: hitobito
+      RAILS_DB_PASSWORD: hitobito
+      RAILS_DB_NAME: hitobito_test
+      RAILS_TEST_DB_NAME: hitobito_test
+
+    services:
+      mysql:
+        image: 'mysql:5.7'
+        env:
+          MYSQL_USER: 'hitobito'
```
BacLuc, line 30: I could not use workflow or job environment variables here.
That's why the db username, pw and other stuff is duplicated here

.github/workflows/tests.yml
```diff
@@ -0,0 +1,87 @@
...
+      memcached:
+        image: 'memcached'
+        ports: [ '11211:11211' ]
+
+    steps:
+      - name: 'Checkout'
+        uses: actions/checkout@v2
+
+      - name: 'Set up Ruby'
+        uses: ruby/setup-ruby@v1.66.0
```
BacLuc, line 50: Using bundle-cache caused problems.
I described them in the Issue #1190

.github/workflows/tests.yml
```diff
@@ -0,0 +1,87 @@
...
+          echo "bundle version: $(bundle -v)"
+          echo "node version: $(node -v)"
+          echo "yarn version: $(yarn -v)"
+
+      - name: 'Bundle install'
+        run: |
+          bundle install --jobs 4 --retry 3
+
+      - name: 'Make changes to Gemfile.lock transparent'
+        run: git diff Gemfile.lock || true
```
BacLuc, line 68: Just to make the changes bundle install makes to the lock file transparent.
There are differences, see: https://github.com/BacLuc/hitobito/actions/runs/628093411

BacLuc:
I will wait for review and then try to retrigger if not already done.

The travis test failed because of an instability:
PeopleController PUT update displays old value again

     Failure/Error: if Settings.assignments&.enabled

       #<Double (anonymous)> was originally created in one example but has leaked into another example and can no longer be used. rspec-mocks' doubles are designed to only last for one example, and you need to create a new one in each example you wish to use it for.

PR #1200, opened 2021-03-08
Fix arc42 link

uroesch:
Summary:
  * Link to the official arc42 documentation was outdated.
    Fix the URL so the link works again.


carlobeltrame (approved):
Thanks!

PR #1234, opened 2021-04-06
Feature/bbt 26 bulk mail

miosidler:
Initial finish of MailRetrieverJob


psunix:
will be finished @hitobito-bbt first

PR #1259, opened 2021-04-27
Feature/30 catch all

miosidler:
Implements Imap client and access to catch-all mailbox for top level users (Settings->Subscription Mails). There mails can be deleted and moved to spam or failed mailbox.


psunix:
wird später gemerged

PR #1268, opened 2021-04-30
Implement IMAP Client

luca-bruegger:
# IMAP Client implementation
## Still Failing Tests
Unit Tests located in the MailingLists Folder within rspec's spec Root Directory are still failing. In cooperation with @psunix and @miosidler, we've already tried to fix this issue. Strangely, if the tests got executed one-by-one they have worked properly.

The reason why they are still failing, could be the caching of the :manage Ability within the various_ability.rb:27 file. Because every single test, throws an

```
CanCan::AccessDenied:
       You are not authorized to access this page.
```

Execption.

# ToDos

- [ ] Review Code
- [ ] Test Client in UI (there's a test mailinglist e-mail account, https://cryptopus.puzzle.ch/accounts/5359)
- [ ] Fix Mocking of Ability in specs

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1,7 +1,7 @@
 PATH
   remote: ../hitobito_generic
   specs:
-    hitobito_generic (1.22.17)
+    hitobito_generic (1.26.0)
 
 GEM
   remote: https://rubygems.org/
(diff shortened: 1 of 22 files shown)


TheWalkingLeek:
The problem sounds like it has to do with when the ability gets included no? 
The `on(Imap::Mail)` call is made based on the settings and as soon as the file gets included (right?). Meaning, if any test prior to this one includes the VariousAbility before the mocking happens, the permissions are simply not registered. Would also explain why it works when running the spec on its own.
Maybe try and pry inside the `on` block and see if and when it gets executed...

psunix:
new PR created

PR #1342, opened 2021-07-05
Repair ability-definition to fix Specs

### LINKED ISSUE #1332: TESTS: IMAP Client Tests fixen
## Aktuelles Verhalten

Tests auf master failen

## Gewünschtes Verhalten

Tests auf master laufen durch

## Schritte zum reproduzieren

https://github.com/hitobito/hitobito

# Tech-Spec

Siehe Diskussion in https://github.com/hitobito/hitobito/pull/1268

kronn:
An ability in a conditional might not be defined, with a qualifier it can always be loaded, yielding consistent results.

Otherwise, the presence of this permission is dependent on the order of
the specs which maybe set the configuration-setting to define the
permission.

I also moved the ability to the other mail-things as I expected it
there.

Fixes #1332

### DIFF
--- a/app/abilities/assignment_ability.rb
+++ b/app/abilities/assignment_ability.rb
@@ -6,7 +6,6 @@
 #  https://github.com/hitobito/hitobito.
 
 class AssignmentAbility < AbilityDsl::Base
-
   on(Assignment) do
     class_side(:index).all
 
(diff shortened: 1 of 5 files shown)


carlobeltrame (approved):
Wow, nice catch and good solution! Thanks!

PR #1399, opened 2021-09-26
Add more fine-grained oauth scopes to access existing APIs

simfeld:


app/domain/event/participation_filter.rb
```diff
@@ -38,10 +37,6 @@ def event
     Event.find_by(id: @event_id)
   end
 
-  def user
-    Person.find_by(id: @user_id)
-  end
-
```
simfeld, line 44: Removed since I didn't see any usages

app/controllers/concerns/authenticatable.rb
```diff
@@ -59,9 +64,12 @@ def prepare_authorization_via_shared_access_token
...
     return unless token&.accessible?
-    return head(:forbidden) unless token.acceptable?(:api)
-
-    sign_in token.person, store: false
+    acceptable = token.acceptable?(:people) ||
+      token.acceptable?(:groups) ||
+      token.acceptable?(:events) ||
+      token.acceptable?(:api)
+    return head(:forbidden) unless acceptable
+    sign_in token, store: false
```
simfeld, line 72: To be verified: Not signing in as `token.person` anymore (but defining the abilities via `DoorkeeperTokenAbility`) might limit the `:api` scope in some places. Currently, reading people, group, event and invoice data is supported. This covers everything currently mentioned in https://github.com/hitobito/hitobito/blob/master/doc/development/05_rest_api.md. Is there anything else?

carlobeltrame: I went through all the serializers in app/serializers and looked for ones that are used in controllers. It's the following list currently:
`AppStatusSerializer` (only used for application health checks), `EventParticipationSerializer`, `EventSerializer`, `GroupListSerializer`, `GroupSerializer`, `InvoiceSerializer`, `ListSerializer` (used together with other serializers for rendering lists instead of single models), `MailingListSerializer`, `PeopleSerializer`, `PersonSerializer`, `UserSerializer` (used for the deprecated User API Token endpoints).
Out of these, I think you are only missing the mailing list endpoints, which haven't been documented yet in 05_rest_api.md.

carlobeltrame (changes requested):
Nice job overall, looks like a rather clean solution to me. I especially like the way you delegate to the user ability once the token is deemed acceptable.
Do you think you could add tests for the other endpoints (events, invoices, mailing lists) as well? And we also need to update the [documentation](https://github.com/hitobito/hitobito/blob/master/doc/development/08_oauth.md) with the new scopes

app/domain/event/participation_filter.rb
simfeld: Reverted and fixed `user_id` logic in 38e5c33

app/controllers/concerns/authenticatable.rb
simfeld: Added mailing list scope in b06e5b8

app/models/oauth/access_token.rb
```diff
@@ -34,6 +34,8 @@ class AccessToken < Doorkeeper::AccessToken
     # no need to check expires_in as refresh_token would never expire
     scope :active, -> { where(revoked_at: nil) }
 
+    devise :timeoutable
```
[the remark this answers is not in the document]
simfeld: 08474db

simfeld:
> Nice job overall, looks like a rather clean solution to me. I especially like the way you delegate to the user ability once the token is deemed acceptable. Do you think you could add tests for the other endpoints (events, invoices, mailing lists) as well? And we also need to update the [documentation](https://github.com/hitobito/hitobito/blob/master/doc/development/08_oauth.md) with the new scopes

I think I addressed all your comments now. Tests are added in 25c351b.

carlobeltrame (approved):
Ich habs nochmals von Hand getestet, und es klappt soweit ich sehe.
Man muss sich aber bewusst sein, dass die Ability vom Token liest, nicht von der OAuth Applikation. Das bedeutet dass ein Token seine Berechtigungen beibehält, auch wenn inzwischen der OAuth Applikation die Rechte weggenommen wurden. Also wenn man eine OAuth Applikation per sofort ganz von einer Berechtigung ausschliessen will, dann muss man zusätzlich auch noch die aktiven OAuth Autorisierungen (= ausgestellte Access Tokens) löschen gehen.

PR #1468, opened 2021-11-08
Fix tests on docker setup, improve setup

severinraez:
Belongs to https://github.com/hitobito/development/pull/31

* Add Wagonfile to allow specifying wagons via env var WAGONS
* Work around docker bug to fix tests
* Don't load wagon twice when testing

Refs https://github.com/hitobito/hitobito/issues/1463


spec/support/work_around_docker_fs_bug.rb
```diff
@@ -0,0 +1,19 @@
+# Workaround for https://github.com/docker/for-linux/issues/1015,
```
mtnstar, line 1: we could also add this file to development setup repo and mount it

mtnstar (approved):
:+1:

PR #1547, opened 2021-12-14
Feature/sbv 104 Import and validate addresses only when configured

TheWalkingLeek:
Refs: https://github.com/hitobito/hitobito_sbv/issues/104


kronn (approved):
Sehr gut.

PR #1571, opened 2021-12-27
Ruby 2.7

kronn:
Beim Deployment muss noch der ImageStream für das base-image von Ruby 2.5 auf Ruby 2.7 umgestellt werden.


mtnstar:
updating docker dev setup (https://github.com/hitobito/development) seams to be straight forward. just update rails docker image source to 2.7

carlobeltrame:
Also bundler must be updated in the Dockerfile: https://github.com/hitobito/development/blob/master/docker/rails.dockerfile#L13

kronn:
> Also bundler must be updated in the Dockerfile: https://github.com/hitobito/development/blob/master/docker/rails.dockerfile#L13

With a ruby27-centos7 image, we get a newer bundler anyway and do not need to update it manually. But you're right, this needs to be adapted (deleted, in this case).

PR #1634, opened 2022-02-14
Bump webpacker

kronn:
This upgrades webpacker a little to address a few security issues. Along the way, I removed some helpers that were not supported anymore or gone. This resulted in the removal of `compass` from the dependencies. Since the npm-package seems unmaintained, this further reduces the dependency-graph.

For PBS, the CSS-output is the same, I did not compare the others. I did make a "compile test" for all wagons I know about. This resulted in the two linked PRs.


carlobeltrame (approved):
LGTM und funktioniert bei mir. Merci!

PR #1686, opened 2022-03-12
Disallow empty person searches

Joelius300:
Before this PR you could search for people in the add person dialogs with just spaces (returns any 10 rows). This probably applies to every controller implementing search via the Searchable module because the parameter from the frontend doesn't have whitespace stripped. This PR fixes this by simply removing leading and trailing whitespace before further processing it.

### Todo
- Do the automated tests still pass?
- Test more potentially affected places (I tested "Person hinzufügen" for Kurse and Lager in the pbs wagon).


carlobeltrame (approved):
Thanks!

PR #1718, opened 2022-04-13
Move from Carrierwave to ActiveStorage

kronn:
See #1568 

This is the part where we migrate from Carrierwave to some ActiveStorage. Right now, this means local-storage. Therefore, it is still a draft PR.

Before merging this, I want to set up S3-buckets for our integrations and productions and have those available for the migration.

When merging this, it needs to be squashed, so we have 1 commit that we can apply as hotfix to all instances.

A preliminary review could be interesting already.


app/helpers/upload_display_helper.rb
```diff
@@ -0,0 +1,87 @@
...
+      model.send(name.to_sym).yield_self do |pic|
+        if size
+          # variant passes to mini_magick or vips, I assume mini_magick here
+          pic.variant(resize_to_limit: extract_image_dimensions(size))
+        else
+          pic
+        end
+      end
+    elsif model.respond_to?(:"carrierwave_#{name}") && model.send(:"carrierwave_#{name}")
+      model.send(:"carrierwave_#{name}_url")
```
carlobeltrame, line 41: If this ever happens on a production env, I as a dev would want to know about it. Maybe you could send a message to sentry, without actually raising an exception?

app/helpers/upload_display_helper.rb
```diff
@@ -0,0 +1,87 @@
...
+      upload_default(default)
+    end
+  end
+
+  # return the filename of the uploaded file
+  def upload_name(model, name)
+    if model.send(name.to_sym).attached?
+      model.send(name.to_sym).filename.to_s
+    elsif model.respond_to?(:"carrierwave_#{name}_identifier")
+      model.send(:"carrierwave_#{name}_identifier")
```
carlobeltrame, line 52: Same here

app/helpers/upload_display_helper.rb
```diff
@@ -0,0 +1,87 @@
...
+    end
+  end
+
+  def upload_exists?(model, name)
+    return true if model.send(name.to_sym).attached?
+
+    if model.respond_to?(:"carrierwave_#{name}")
+      model.send(:"carrierwave_#{name}").present?
```
carlobeltrame, line 60: Same here

app/helpers/upload_display_helper.rb
```diff
@@ -0,0 +1,87 @@
...
+    end
+  end
+
+  private
+
+  def upload_variant(model, name, variant, default: model.name.underscore)
+    if model.send(name.to_sym).attached?
+      model.send(name.to_sym).variant(variant.to_sym)
+    elsif model.respond_to?(:"carrierwave_#{name}")
+      model.send(:"carrierwave_#{name}").send(variant.to_sym).url
```
carlobeltrame, line 72: And here

app/helpers/upload_display_helper.rb
```diff
@@ -0,0 +1,87 @@
...
+    else
+      upload_default([default, variant].compact.map(&:to_s).join('_'))
+    end
+  end
+
+  def upload_default(png_name = 'profil')
+    ActionController::Base.helpers.asset_pack_path("media/images/#{png_name}.png")
```
carlobeltrame, line 79: My security alarm bells are ringing, when you pass unknown input into a file path, input which could easily be provided by the user in some wagon. But I think it's fine here because the path will be looked up in the webpacker manifest. So effectively, we have an allowlist in place here.

config/application.rb
```diff
@@ -73,6 +73,8 @@ class Application < Rails::Application
     config.cache_store = :mem_cache_store, { compress: true,
                                              namespace: ENV['RAILS_HOST_NAME'] || 'hitobito' }
 
+    config.active_storage.variant_processor = :mini_magick
```
carlobeltrame, line 76: We have to make sure minimagick is installed in the deployed docker images as well as in the development setup then

app/helpers/upload_display_helper.rb
kronn: Hmm. this helper is transitional in nature and not intended to be there after the data-migration is done. We have #1658 to clean this up. I will think a little if adding notifications is easily possible (with detection of wether the migration is done already).

app/helpers/upload_display_helper.rb
kronn: Very good point. I missed the possibility/risk of using this in a wagon. I will try to mitigate this. FWIW, this happens already in the currently present Uploaders, with the same confinment to the webpacker-realm.

config/application.rb
kronn: We already use mini_magick for image processsing.

app/helpers/upload_display_helper.rb
```diff
@@ -76,6 +76,8 @@ def upload_variant(model, name, variant, default: model.name.underscore)
   end
 
   def upload_default(png_name = 'profil')
+    png_name = 'profil' if png_name.include?('..')
```
carlobeltrame, line 79: This unfortunately helps only marginally. See https://github.com/xmendez/wfuzz/blob/master/wordlist/vulns/dirTraversal-nix.txt for a big list of ideas how to do path traversal without two consecutive periods.

kronn: Agreed. In theory, at least.

Practically, though, I cannot read even allowed files with path-traversal

`media/images/profil.png` gives me an internal path to a PNG
`media/images/../images/profile.png` gives me nothing but a Webpacker-Error
`media/images/%2e%2e/images/profile.png` gives me nothing but a Webpacker-Error

Since this is developer-provided, non-user-facing and seemingly secured already, I tend to not fix it. If a dev wants to be clever but is not, it now defaults to the default profile-picture.

TheWalkingLeek (approved):

TheWalkingLeek (approved):
Ich mag das :+1:

spec/domain/export/pdf/messages/letter/header_spec.rb
```diff
@@ -58,91 +67,94 @@ def expects_image(id)
...
-      let(:image) { @image }
-      let(:image_group_id) { assign_image(top_group) }
-      let(:image_path) { %r{/picture/#{image_group_id}/logo.*\.png} }
+      let(:image) { Rails.root.join("spec/fixtures/files/#{@image}") }
+      # let(:image_group_id) { assign_image(top_group) }
+      # let(:image_path) { %r{/picture/#{image_group_id}/logo.*\.png} }
 
-      it 'does not scale if image smaller than logo box' do
-        @image = fixture_file_upload('images/logo.png') # 230x30px
+      xit 'does not scale if image smaller than logo box' do
```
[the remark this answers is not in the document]
kronn: Anpassen wäre das lange Ziel. Wir machen das Skalieren noch, den passenden Code habe ich auf ActiveStorage umgeschrieben. Bei der Spec habe ich das Gefühl, dass das eher ein Cleanup-Task ist. Ich bin schon viel zu lange an dem Thema...

app/jobs/download_cleaner_job.rb
```diff
@@ -16,27 +16,15 @@ def perform_internal
   private
 
   def remove_old_downloads
-    removable_files.each do |file|
-      File.delete(file)
-    end
-  end
+    AsyncDownloadFile.where(older_than_a_day).find_each do |file|
+      next unless file.generated_file.attached?
```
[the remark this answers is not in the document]
kronn: ja, das entferne ich

PR #1739, opened 2022-05-12
Add fields to define role created_at and deleted_at manually

TheWalkingLeek:
Closes: #1714 

Merge with:
- https://github.com/hitobito/hitobito_pbs/pull/226
- https://github.com/hitobito/hitobito_sww/pull/44
- https://github.com/hitobito/hitobito_sbv/pull/118


mtnstar:
looks pretty good :tada: 

* make sure permitted attrs only allow deleted_at if current user is allowed to delete role
  * I've already committed some code (WIP)
* instead of hiding, disable deleted_at field if user is not allowed to delete record

mtnstar (changes requested):
see comment

PR #1822, opened 2022-08-27
Add local noto-sans font

olibrian:
Add noto-sans font local and replace import from google api
fix https://github.com/hitobito/hitobito/issues/1632


carlobeltrame (approved):
The font is customized in the following wagons:
- glp, sbv, skv, insieme, die_mitte, dsj, svse, sjas, jubla: `Helvetica, Arial, sans-serif;` no google fonts in use
- sww: `"PT Sans", "Helvetica Neue", Helvetica, Arial, sans-serif;` not loaded from google
- [ ] pbs: `'Source Sans Pro', sans-serif;` :arrow_right: same fix should be added to the PBS wagon

PR #1883, opened 2022-10-13
Feature/1857 truemail random invalid

bihorco36:


spec/models/person_spec.rb
```diff
@@ -567,6 +567,23 @@ def should_not_be_valid_swiss_post_code
       expect(person.errors.messages[:email].first).to eq('ist nicht gültig')
     end
 
+    it 'sends sentry notification if correct email is invalid' do
+      allow(Truemail).to receive(:valid?).and_return(false)
+      person.email = 'dude@blabliblablu.ch'
```
mtnstar, line 572: going to reserve this domain right away :joy:

spec/domain/app_status/mail_spec.rb
```diff
@@ -79,6 +80,30 @@
...
+      expect(imap_connector)
+        .to receive(:fetch_mails)
+        .with(:inbox)
+        .and_return([])
+    end
+
+    it "doesn't verify correctly anymore" do
+      allow(Truemail).to receive(:valid?).and_return(false)
+
+      expect(app_status.code).to eq(:ok)
```
mtnstar, line 95: should this still have status code :ok if true mail failing ?

bihorco36: I wasn't sure about that either. In theory, the mail server is still working and only the saving of a person is impaired, therefore the :service_unavailable status would be wrong. But we might could implement another fitting status if only Truemail would fail. Do you have any suggestions for a good matching status?

app/domain/app_status/mail.rb
```diff
@@ -16,10 +16,14 @@ class AppStatus::Mail < AppStatus
...
   def initialize
     @catch_all_inbox = catch_all_inbox
+    @truemail_working = truemail_working
   end
 
   def details
-    { catch_all_inbox: @catch_all_inbox }
+    {
+      catch_all_inbox: @catch_all_inbox,
+      truemail_working: @truemail_working
```
mtnstar, line 25: we want to use this with ocp4 health checks.
shouldn't we add this to default health check instead of mail specific?

PR #1935, opened 2022-12-03
Do not run when the PR is opened/reopened, just on push

kronn:
We are only interested to run against code-changes. If this happens to a Branch that ist part of a PR or not is not relevant. Furthermore, the push-trigger was there all the time, so the 'PR has opened'-trigger is redundant and most of the time late.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -5,6 +5,12 @@ on:
     paths-ignore:
       - 'doc/**'
       - '**.md'
+    branches:
+      - 'master'
+  pull_request:
+    paths-ignore:
+      - 'doc/**'
+      - '**.md'
 
 jobs:
   build:
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -5,10 +5,6 @@ on:
     paths-ignore:
       - 'doc/**'
       - '**.md'
-  pull_request:
-    paths-ignore:
-      - 'doc/**'
-      - '**.md'
 
 jobs:
   build:


carlobeltrame (changes requested):
I disagree. The pull request event triggers on a simulated merged version of the code. So this protects us from CI failures which only happen after merging (because of flaws in the automatic git merge algorithm). So for PRs, I'd argue that the pull_request trigger is more important and realistic than the push trigger.

kronn:
Conceptually, I agree. That would be nice and of course be quite different. From what I gathered from the docs, however, the trigger just starts a job without knowledge what to do later on.

If the current tests-workflow does indeed behave differently in a PR than on master, I could not see it, but it should also still behave like that.

If that simulated merge would be made, then it would also require a rerun every time the respective merge-target is updated. This does not seem to be the case (as it would cost quite a lot of computation-power to ensure ongoing "good to merge"-status).

If you feel that the `pull_request`-trigger makes sense, we could also restrict the `push`-trigger to only cover the master branch. This would however result in a later start of a workflow ("after the PR has been opened" vs "after I pushed the code").

Restricting to "push on master" and "pull_request opened/reopened" would also result in fewer builds, which is fine by me as well. I am unsure if this would still include code-updates to the PR, but this is something we will find out rather quickly.

In conclusion, I think that "push" without restrictions is the simplest trigger without sacrificing functionality. Am I missing something?

carlobeltrame:
https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request
> Note that GITHUB_SHA for this event is the last merge commit of the pull request merge branch. If you want to get the commit ID for the last commit to the head branch of the pull request, use github.event.pull_request.head.sha instead.

It is really something different. Many GitHub Actions, including the extremely common [checkout action](https://github.com/actions/checkout) will use the `GITHUB_SHA` by default and thus work differently on push and pull_request events.
I agree it is kind of inconsistent of GitHub to not re-run the PR CI when the merge base changes, but that is probably a measure GitHub already takes to avoid too many unnecessary CI runs. I also see the relatively new note "Workflows will not run on pull_request activity if the pull request has a merge conflict." This also points towards optimizations GitHub already does.

Why are you trying to reduce the number of runs on our side at all? Saving energy? It does not cost us anything to run on both `push` and `pull_request`, and it's common practice in open source projects to always run CI at least on the pull_request event: [[1]](https://github.com/keycloak/keycloak/pull/15822) [[2]](https://github.com/puppeteer/puppeteer/blob/main/.github/workflows/ci.yml) [[3]](https://github.com/elixir-lang/elixir/blob/main/.github/workflows/ci.yml) [[4]](https://github.com/actions/checkout/blob/main/.github/workflows/test.yml) [[5]](https://github.com/rails/rails/blob/main/.github/workflows/rubocop.yml). Special case VSCode: [only manual CI runs](https://github.com/microsoft/vscode/pull/149246/commits/776233f0ffd672095f8b5d7ee0af65330944654c)
(Note: It's kind of hard to find big Open Source projects which work with PRs from the origin, as opposed to only forks. But many of them have dependabot or similar which pushes directly to the origin.)

kronn:
If that is indeed something different (which I did not gather from the
docs), thanks for clarifying, @carlobetrame.

On Mon, Dec 05, 2022 at 02:09:56AM -0800, Carlo Beltrame wrote:
> I also see the relatively new note "Workflows will not run on pull_request activity if the pull request has a merge conflict."

This however reads to me that (like now), the action would run on
general, mergable activity on the PR. Therefore a separate push-trigger
is not needed. The merge-conflict needs to be resolved anyway before
merging, which will result in activity, triggering another PR-build.

> Why are you trying to reduce the number of runs on our side at all? Saving energy?

Mostly saving mental overhead. Two checks that look like they do the
same thing are confusing to me. The same thing here is "is the code
still valid according to our specs".

The other two checks are "have secrets been exposed" and "did we open
new or newly found security holes".

> It does not cost us anything to run on both `push` and `pull_request`, and it's common practice in open source projects to always run CI at least on the pull_request event [...]

Still does not make it free, but money/energy was not my main concern.

Since the consensus seems to be to keep the `pull_request`-trigger (fine
by me), I will update the PR to keep that trigger for PRs but restrict
builds on push to the master-branch.

If there is no PR, there is no point to let CI run. The obvious
exclusion are pushes to master, of course.

This should result in the check stating "the specs will pass after a
merge". The more esoteric "the specs are fine in this branch, but not
after a merge" is interesting, but effectively useless, therefore waste.

Cheers,
Matthias

carlobeltrame (approved):

PR #1946, opened 2022-12-13
Really exclude archived roles from mailing lists

carlobeltrame:
Fixes https://github.com/hitobito/hitobito/issues/1440#issuecomment-1348178994

It is not enough to delete all subscriptions of a group while archiving, because the roles of the group might also be included in the mailing list via a subscription of a higher group. E.g. say an Abteilung has 3 child groups with the same type, and the same "Member" role. If on the Abteilung there is a mailing list including all "Member" roles in the Abteilung, then I can archive one of the child groups and deleting its subscriptions will do nothing. The actual subscription belongs to the Abteilung, so it looks something like this:
```
Subscription {
  group: Abteilung,
  role_types: ['Group::ChildGroup::Member'],
}
```

For this reason, we really have to explicitly exclude archived roles from all mailing lists.


TheWalkingLeek (approved):
Makes sense, thanks for fixing! :)

PR #1989, opened 2023-01-22
Allow for integration of nextcloud

### LINKED ISSUE #1854: OAuth: NextCloud Login mit Hitobito ermöglichen
Als Hitobito-Nutzer, der auch Nextcloud nutzt
möchte ich mich bei NextCloud via hitobito einloggen
und Rolleninformationen übergeben können,
um die Berechtigung für alle meine Vereinsmitglieder konsistent verwalten zu können.

NextCloud kann mit dem [OpenID User Backend](https://github.com/nextcloud/user_oidc) die Authentication via OAuth z.B. gegen hitobito vornehmen.

Idealerweise kann man die hitobito-Rollen NextCloud so mitteilen, dass daraus NextCloud-Gruppen mit entsprechenden Berechtigungen werden. 

# Tech-Spec

- Ziel ist die zentrale Nutzer- und Berechtigungsverwaltung
- schöner Nebeneffekt ist, dass man nur noch seinen hitobito-Login braucht
- die initiale Verbindung (Providerconfig in hitobito und NextCloud, Gruppen in NextCloud) erfolgt manuell bis es so oft passiert, dass wir es dokumentieren und später automatisieren können.

# ToDo

- [x] NextCloud-Testinstanz bereitstellen, um Integration zu validieren
- [x] Möglichkeit finden, um hitobito-Rollen an NextCloud weiterzugeben. (`user_oidc` mit PR 502)
- [x] Rollen-Klassen erweitern, um Gruppenmapping zu unterstützen (ein bestimmte Rolle innerhalb einer bestimmten Ebene bekommt direkt Gruppenname zugewiesen)
- [x] Gruppen erweitern, um Link zu NextCloud aufzunehmen (String `nextcloud_url` auf `groups`) (nur Layergruppen)
- [x] OAuth-Scope hinzufügen, um die für NextCloud notwendigen Daten zurückzugeben.
- [x] Doku hinzufügen, um tatsächliche Integration festzuhalten. 
- [x] Specs schreiben
- [x] Mit angemessener Rolle "durchklicken"
- [x] CHANGELOG-Eintrag unter "unreleased" unten hinzufügen

kronn:
Everything is hidden behind the FeatureGate `groups.nextcloud`. This is disabled by default.

Once activated, you can

- add a nextcloud-url to any layer-group and have it displayed on every group in the layer
- use a nextcloud-scope for oauth-apps. This allows to pass user information and a list of groups to nextcloud to provision users there.
- configure roles to either map to a fixed groupname or a dynamic group, depending on the group you have the role on.

Apart from that, I added a rake-task to determine the correct configuration quicker. The idea is to have checks for everything that requires extra-steps to be configured, collected in `rake hitobito:check_config`.

Closes #1854

### DIFF
--- a/app/decorators/group_decorator.rb
+++ b/app/decorators/group_decorator.rb
@@ -97,14 +97,17 @@ class GroupDecorator < ApplicationDecorator
                            .pluck(:id)
   end
 
-  def nextcloud_url
-    layer_group.nextcloud_url
-  end
-
   def archived_class
     return nil unless model.archived?
 
     'is-archived'
+  end
+
+  def nextcloud_organizer
+    model
+      .hierarchy.where(layer_group_id: layer_group.id) # local_hierarchy as a scope
+      .where.not(nextcloud_url: nil) # only those with a nextcloud_url
+      .last
   end
 
   private
--- a/app/decorators/group_decorator.rb
+++ b/app/decorators/group_decorator.rb
@@ -1,6 +1,6 @@
 # frozen_string_literal: true
 
-#  Copyright (c) 2012-2021, Jungwacht Blauring Schweiz. This file is part of
+#  Copyright (c) 2012-2023, Jungwacht Blauring Schweiz. This file is part of
 #  hitobito and licensed under the Affero General Public License version 3
 #  or later. See the COPYING file at the top-level directory or at
 #  https://github.com/hitobito/hitobito.
@@ -95,6 +95,10 @@ class GroupDecorator < ApplicationDecorator
     @subgroup_ids ||= Group.where('lft >= :lft AND rgt <= :rgt',
                                   lft: group.lft, rgt: group.rgt)
                            .pluck(:id)
+  end
+
+  def nextcloud_url
+    layer_group.nextcloud_url
   end
 
   def archived_class
--- a/app/models/role.rb
+++ b/app/models/role.rb
@@ -163,7 +163,7 @@ class Role < ActiveRecord::Base
   # If this role has contact_data permissions, set the flag on the person
   def set_contact_data_visible
     if becomes(type.constantize).permissions.include?(:contact_data)
-      person.update_column :contact_data_visible, true
+      person.update_column :contact_data_visible, true # rubocop:disable Rails/SkipsModelValidations intentional
     end
   end
 
@@ -171,20 +171,20 @@ class Role < ActiveRecord::Base
   def reset_contact_data_visible
     if permissions.include?(:contact_data) &&
        !person.roles.collect(&:permissions).flatten.include?(:contact_data)
-      person.update_column :contact_data_visible, false
+      person.update_column :contact_data_visible, false # rubocop:disable Rails/SkipsModelValidations intentional
     end
   end
 
   def set_first_primary_group
     if (deleted_at.nil? || deleted_at.future?) && person.roles.count <= 1
-      person.update_column :primary_group_id, group_id
+      person.update_column :primary_group_id, group_id # rubocop:disable Rails/SkipsModelValidations intentional
     end
   end
 
   def reset_primary_group
     if person.primary_group_id == group_id &&
         person.roles.where(group_id: group_id).count.zero?
-      person.update_column :primary_group_id, alternative_primary_group.try(:id)
+      person.update_column :primary_group_id, alternative_primary_group.try(:id) # rubocop:disable Rails/SkipsModelValidations intentional
     end
   end
 
--- a/app/models/role.rb
+++ b/app/models/role.rb
@@ -70,6 +70,12 @@ class Role < ActiveRecord::Base
     # { 'gid' => 'group-id-that-is-unique-in-nextcloud', 'displayName' => 'Name of Group' }
     class_attribute :nextcloud_group
     self.nextcloud_group = false
+
+    NextcloudGroup = Struct.new(:gid, :displayName) do
+      def hash
+        @gid.hash
+      end
+    end
   end
 
   # If these attributes should change, create a new role instance instead.
@@ -150,6 +156,21 @@ class Role < ActiveRecord::Base
   def nextcloud_group
     FeatureGate.assert!('groups.nextcloud')
 
+    info = nextcloud_group_details
+
+    return if info.nil?
+
+    case info
+    when String then NextcloudGroup.new(info, info)
+    when Hash then NextcloudGroup.new(info['gid'], info['displayName'])
+    end
+  end
+
+  private
+
+  def nextcloud_group_details
+    return nil unless FeatureGate.enabled?('groups.nextcloud')
+
     case (setting = self.class.nextcloud_group)
     when String then { 'gid' => "hitobito-#{setting}", 'displayName' => setting }
     when true   then { 'gid' => group_id.to_s,         'displayName' => group.name }
@@ -157,8 +178,6 @@ class Role < ActiveRecord::Base
     when Proc   then setting.call(self)
     end
   end
-
-  private
 
   # If this role has contact_data permissions, set the flag on the person
   def set_contact_data_visible
--- a/app/models/role.rb
+++ b/app/models/role.rb
@@ -55,9 +55,19 @@ class Role < ActiveRecord::Base
   self.merge_excluded_attributes = []
 
   FeatureGate.if('groups.nextcloud') do
-    # Can be either a string which is simply returned or a boolean.
-    # If true, then the attached Group should be referenced
-    # If false, no group for nextcloud is referenced
+    # Can be one of several types:
+    #
+    # String - Name of the nextcloud-group
+    # String  - Name of the nextcloud-group
+    # Boolean - Dynamic lookup of the nextcloud-group
+    #           If true, then the attached Group should be referenced
+    #           If false, no group for nextcloud is referenced
+    # Symbol  - Identifier of an instance-method
+    # Proc    - A proc that is called with the role
+    #
+    # Both Symbol and Proc are expected return something that is useful to Nextcloud.
+    # It can either only return the name of the nextcloud-group or a Hash like this
+    # { 'gid' => 'group-id-that-is-unique-in-nextcloud', 'displayName' => 'Name of Group' }
     class_attribute :nextcloud_group
     self.nextcloud_group = false
   end
@@ -138,11 +148,13 @@ class Role < ActiveRecord::Base
   end
 
   def nextcloud_group
-    FeatureGate.assert! 'groups.nextcloud'
+    FeatureGate.assert!('groups.nextcloud')
 
     case (setting = self.class.nextcloud_group)
     when String then { 'gid' => "hitobito-#{setting}", 'displayName' => setting }
     when true   then { 'gid' => group_id.to_s,         'displayName' => group.name }
+    when Symbol then method(setting).call
+    when Proc   then setting.call(self)
     end
   end
 
--- a/app/models/role.rb
+++ b/app/models/role.rb
@@ -71,9 +71,13 @@ class Role < ActiveRecord::Base
     class_attribute :nextcloud_group
     self.nextcloud_group = false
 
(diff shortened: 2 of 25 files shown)


Michael-Schaer:
Das sieht spannend aus!

Wenn es da Möglichkeiten gibt, das bei uns zu aktivieren, haben wir die Ohren offen :)

db/migrate/20230115160409_add_nextcloud_link_to_groups.rb
```diff
@@ -0,0 +1,5 @@
+class AddNextcloudLinkToGroups < ActiveRecord::Migration[6.1]
```
TheWalkingLeek, line 1: license

TheWalkingLeek (approved):
Looks good to me, maybe someone with more Oauth knowledge could look over the Oauth parts just to be sure :)

app/models/role.rb
```diff
@@ -129,6 +137,15 @@ def archived?
     archived_at.present?
   end
 
+  def nextcloud_group
+    FeatureGate.assert! 'groups.nextcloud'
+
+    case (setting = self.class.nextcloud_group)
+    when String then { 'gid' => "hitobito-#{setting}", 'displayName' => setting }
```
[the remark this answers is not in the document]
kronn: As far as I understand it `gid` is the group id. I could also only pass the names and the nextcloud-plugin that does the mapping generates a sha256 from the name to have a unique ID. Since the group-names can change, I prefer this.

db/migrate/20230115160409_add_nextcloud_link_to_groups.rb
kronn: oops. updated.

kronn:
I'd like your OAuth-expertise on this one, @carlobeltrame. Could you review this?

app/models/role.rb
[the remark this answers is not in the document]
carlobeltrame: Do you intend to serve two different use cases with the boolean vs. string? If I specify a string (e.g. "Trainers" in SKV), and there are more than one group of that type in hitobito, all people from all these groups will go to the same nextcloud group. On the other hand, if I specify just `true`, all people from these different trainer groups will be put into separate nextcloud groups. Seems confusing to me as a developer: usually `true` is just a shorthand for the default behaviour in such DSLs, but here there is no way to acheive with a string the same behaviour as `true` produces.
If both ways are definitely needed, how about letting the role class define a method or Proc class attr instead, which can generate the nextcloud group id and display name based on the role instance? This would also be more extensible for the future, if we e.g. ever need to put the user into multiple nextcloud groups based on just one hitobito role.

app/views/groups/_attrs.html.haml
```diff
@@ -13,6 +13,8 @@
       = render_attrs(entry, :self_registration_link, :self_registration_notification_email)
     = render_attrs(entry, :type_name)
     = render_present_attrs(entry, *entry.used_attributes(:description))
+    - if FeatureGate.enabled?('groups.nextcloud')
+      = render_present_attrs(entry, :nextcloud_url)
```
carlobeltrame, line 17: Wouldn't our `used_attributes` mechanism also apply here? Couldn't you add the nextcloud_url to the group's used_attrs only if the feature gate allows it? Are there no downsides to nextcloud_url not being in the used_attributes?

config/initializers/doorkeeper_openid_connect.rb
```diff
@@ -60,7 +67,7 @@
   # end
 
   claims do
-    claim(:email, scope: :email)     { |resource_owner| resource_owner.email }
+    claim(:email, scope: :email, response: [:user_info, :id_token]) { |resource_owner| resource_owner.email }
```
carlobeltrame, line 70: Why is this change necessary? Shouldn't the user_info and id_token be linked to the `openid` scope instead?

config/locales/models.de.yml
```diff
@@ -1107,6 +1108,7 @@ de:
       people: Lesen aller Personen, auf die du Zugriff hast
       invoices: Lesen aller Rechnungen, auf die du Zugriff hast
       mailing_lists: Lesen aller Abos, auf die du Zugriff hast
+      nextcloud: Lesen der eingestellten Nextcloud-Gruppen und des Anzeigenamens
```
carlobeltrame, line 1111:
```suggestion
      nextcloud: Lesen deiner Nextcloud-Gruppen und deines Anzeigenamens
```
Aus User-Sicht sind die Nextcloud-Gruppen keine Einstellungsmöglichkeit.

config/settings/test.yml
```diff
@@ -20,3 +20,7 @@ email:
   sender: "hitobito <noreply@localhost>"
   mass_recipient: "hitobito <noreply@localhost>"
   list_domain: localhost
+
+groups:
+  nextcloud:
+    enabled: true
```
carlobeltrame, line 26: Remove? Or do tests not inherit the default settings?

doc/development/08_oauth.md
```diff
@@ -30,6 +30,7 @@ Different external applications need different types of data about their users.
 | `people`        | people API endpoints                      |
 | `invoices`      | invoice API endpoints                     |
 | `mailing_lists` | mailing list API endpoints                |
+| `nextcloud`     | maps roles to nextcloud-groups            |
```
carlobeltrame, line 33:
```suggestion
| `nextcloud`     | nextcloud groups based on user's roles    |
```
A mapping would intuitively mean a hash to me, but this scope just outputs a flat list of groups.

lib/tasks/hitobito.rake
```diff
@@ -34,4 +34,28 @@ namespace :hitobito do
...
+  namespace :checks do
+    task oauth: :environment do
+      signing_key = Settings.oidc.signing_key.join.presence
+
+      if signing_key.nil?
+        puts <<~MESSAGE
+          ❌ OAuth not correctly configured
+
+            JWT Signing Key missing.
+            This key is needed to OAuth to work.
```
carlobeltrame, line 52:
```suggestion
            This key is needed for OAuth to work.
```

app/models/role.rb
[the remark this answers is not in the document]
kronn: Currently, there are the two use-cases you identified. For the SKV we have several Fachkommissionen, which should map to different nextcloud-groups.

The current method could be extended to support a Symbol (pointing to a method) or a Proc (which could be called). I will try if this is as easy as I think it is.

config/settings/test.yml
kronn: The value here is different from the default settings. 

The default is to have it **disabled**, but for tests I want it **enabled**.

config/initializers/doorkeeper_openid_connect.rb
kronn: The default for `response` is `[:user_info]`. This omits the email from the generated id-token. In order to fully provision the user in nextcloud (and allow for mail-notifications), the email needs to passed in the id-token. 

I did not find a way to have one claim have different response-types depending on the scope. Also, the OpenID-scope does not pass the email.

What am I missing here?

app/models/role.rb
```diff
@@ -54,6 +54,14 @@ class Role < ActiveRecord::Base
   class_attribute :merge_excluded_attributes
   self.merge_excluded_attributes = []
 
+  FeatureGate.if('groups.nextcloud') do
```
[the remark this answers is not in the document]
kronn: Good point.

app/views/groups/_attrs.html.haml
kronn: Good point, I'll try that.

app/models/role.rb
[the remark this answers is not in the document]
kronn: I was not able to extract this into a concern, despite following the usual pattern. Since it does not bother me too much personally, I will not do this now.

app/models/role.rb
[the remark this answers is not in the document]
kronn: It was as easy, though I am unsure what "can of worms" I opened with this flexibility. It's not a pandoras box, but can break the integration in a very subtle and hard-to-debug way. I tried to document the class_attribute better, in order to at least provide some help.

app/decorators/group_decorator.rb
```diff
@@ -97,6 +97,10 @@ def subgroup_ids
                            .pluck(:id)
   end
 
+  def nextcloud_url
+    layer_group.nextcloud_url
```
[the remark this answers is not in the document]
kronn: I thought about it and implemented it.

app/views/groups/_attrs.html.haml
kronn: The views are now free from this specific FeatureGate.

I think that in most cases we can and should restrict the FeatureGate to ruby-files.

Good catch.

config/initializers/doorkeeper_openid_connect.rb
kronn: I could try and evaluate the scope associated with the access_token, but Doorkeeper only selects the claims which belong to the scopes associated with the access_token. I could then try an see with other scopes (and therefore claims) are present and determine further steps. 

Each claim can only be linked to one scope (at least in Doorkeeper/OIDC). Linking a central piece of the user to a not always used scope is not useful to me.

OTOH, the id_token only contains data from the claims are covered by the authorized scopes. So, if you do not want to pass your email to your OIDC-app, the scope "email" should not be part of the OAuth-Application. If the consuming app need the e-mail, it will be in the application and (hopefully) authorized by the user. With this, the consuming application receives the data directly with the JWT instead of having to ask for it via the user_info-endpoint.

So, in summary, I think the change is necessary because some OAuth-consumers only evaluate the id_token, especially nextcloud. Since no additional claim is linked to the openid-scope, there is no response to configure.

carlobeltrame (approved):

carlobeltrame (commented):
Mir ist gerade noch aufgefallen: Angenommen ein Layer A und ein Layer B haben beide eine eigene Nextcloud-Instanz, und ich bin Adressverwalter o.ä. im Layer B. Kann ich mich dann im Nextcloud von Layer A einloggen und werde dort als Adressverwalter (entsprechend mit sehr hohen Berechtigungen)? Oder haben wir irgendeinen Mechanismus um das einzuschränken? Bisher gab es eine solche Möglichkeit eben nicht, und jetzt wo wir die Leute animieren, auch höhere Rollen im Nextcloud zu vergeben, wäre das schon noch wichtig. Siehst du das anders?

kronn:
> Mir ist gerade noch aufgefallen: Angenommen ein Layer A und ein Layer B haben beide eine eigene Nextcloud-Instanz, und ich bin Adressverwalter o.ä. im Layer B. Kann ich mich dann im Nextcloud von Layer A einloggen und werde dort als Adressverwalter (entsprechend mit sehr hohen Berechtigungen)? Oder haben wir irgendeinen Mechanismus um das einzuschränken? Bisher gab es eine solche Möglichkeit eben nicht, und jetzt wo wir die Leute animieren, auch höhere Rollen im Nextcloud zu vergeben, wäre das schon noch wichtig. Siehst du das anders?

Es gibt die Möglichkeit zur Einschränkung nicht. Das OAuth-Token weiss zwar zu welcher Person es gehört, aber die zugehörige OAuth-Applikation ist global und nicht an einen Layer gebunden. Man könnte jetzt versuchen, über die Nextcloud-URL und die Redirect-URLs einen Match herzustellen, aber das erscheint mir fehlerträchtig.

Um diese Einschränkung zu ermöglichen, müssten wir OAuth-Applikationen an einen Layer oder eine Gruppe binden. Dies würde ich aber gerne in einem eigenen Schritt/Issue/Ticket machen. Dann kann man auch die Anforderungen besser diskutieren und hat schonmal die Grundlage einer Integration.

Grundsätzlich werden nur Gruppenzugehörigkeiten vergeben. Wenn mehrere gleiche Rollen in Nextcloud unterschiedlich dargestellt werden sollen, müsste man aktuell ein Proc oder so definieren, der klar verschieden Gruppen erstellt.

PR #1993, opened 2023-01-24
Adding UTF8 BOM to CSV-Exports

Michael-Schaer:
Adding an UTF8 BOM allows MS Excel to read UTF-8 correctly whithout changing the charset manually.

The BOM is only added if the settings.yml file explicitly specifies to use UTF8 for csv exports.


Michael-Schaer:
See also https://github.com/hitobito/hitobito_pbs/pull/250

kronn (approved):

PR #2004, opened 2023-02-08
Repair Mailchimp-Sync

kronn:
After a few linting and cleanup steps, I unexpectedly found the bug. 

The perform-methods of the `ReoccuringJob` handles the rescheduling. Therefore, overwriting it prevents the rescheduling.

The current healthz-checks for the DJ pod catch this, but it would be cleaner to reschedule the job directly instead of relying on external reboots.


kronn:
Hmm, the specs need adaptation, unsurprisingly. I'll tackle that at some point when waiting for something else.

PR #2005, opened 2023-02-09
Assign active state during save entry to show correct flash message

TheWalkingLeek:
Fixes: https://github.com/hitobito/hitobito_sww/issues/99

Still not sure about "Anlassverwalter".. There might be better ideas but I wasn't able to find one so I'm tossing it into the room here :)


carlobeltrame (commented):
How about "verantwortliche Personen"?

PR #2012, opened 2023-02-20
Upgrade prometheus exporter

kronn:
In order to get more metrics in our updated dashboards, we need to update the exporter gem. The config did not need changes.


daniel-illi (commented):
:+1:

PR #2047, opened 2023-03-25
Invoice lists: display all invoices by default and make ends of range inclusive

lukas-buergi:
Behaviour before: Invoice lists have the current year (excluding first and last day) preselect, so when looking at old invoice lists, the list is empty and the fact that the end days are not including is surprising.

Behaviour after: Invoice lists by default show invoices starting at the beginning of the year when the list was created, to the end of the current year. The first and last day of the range is included in both the default range and when a different range is chosen by the user.

Tests: A few tests failed in my local run, but they don't look related to this change.

### DIFF
--- a/app/controllers/invoices_controller.rb
+++ b/app/controllers/invoices_controller.rb
@@ -6,7 +6,6 @@
 #  https://github.com/hitobito/hitobito.
 
 class InvoicesController < CrudController
-  include YearBasedPaging
   include Api::JsonPaging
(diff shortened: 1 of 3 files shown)


lukas-buergi:
Should I force-push so the small fix isn't a commit on its own?

carlobeltrame (approved):

carlobeltrame:
Thank you very much for the contribution!

PR #2048, opened 2023-03-25
Dropdown only visible if permission to do action

philobr:
The drop-down menu for modifying a group was visible even though the user did not have permission for any of these actions. Now it is visible only if the user has the permission.

### DIFF
--- a/app/helpers/dropdown/group_edit.rb
+++ b/app/helpers/dropdown/group_edit.rb
@@ -24,8 +24,11 @@ module Dropdown
     private
 
     def init_items # rubocop:disable Metrics/AbcSize,Metrics/MethodLength
-      unless group.archived?
+      unless group.archived? || merge_candidates?
         add_item(translate(:merge), template.merge_group_path(group))
+      end
+
+      unless group.archived? || move_candidates?
         add_item(translate(:move), template.move_group_path(group))
       end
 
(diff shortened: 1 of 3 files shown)


app/helpers/dropdown/group_edit.rb
```diff
@@ -24,8 +24,11 @@ def initialize(template, group)
...
 
     def init_items # rubocop:disable Metrics/AbcSize,Metrics/MethodLength
-      unless group.archived?
+      unless group.archived? || merge_candidates?
         add_item(translate(:merge), template.merge_group_path(group))
+      end
+
+      unless group.archived? || move_candidates?
         add_item(translate(:move), template.move_group_path(group))
       end
```
carlobeltrame, line 33: I find complex `unless` constructs rather hard to read. Could you refactor it so that it gets more readable? E.g.:
```suggestion
      unless group.archived?
        if merge_candidates?
          add_item(translate(:merge), template.merge_group_path(group))
        end

        if move_candidates?
          add_item(translate(:move), template.move_group_path(group))
        end
      end
```
Actually, I think you misread it yourself and introduced a bug. Let me demonstrate:
```ruby
unless group.archived? || move_candidates?
# as per definition of "unless":
if !(group.archived? || move_candidates?)
# apply de Morgan's Theorem
if (!group.archived?) && (!move_candidates?)
```
Thus, in your implementation, the move option is only shown for non-archived groups which *do not have any move candidates*, which seems wrong to me.

The same goes for the move candidates.

carlobeltrame (changes requested):
For this kind of logic change, I think automated tests would be useful to make sure this does not break again in the future. A feature spec would be a good option for this I think, because in a feature spec, a real headless browser is used to access the site, and in the test you can run commands such as `find('.my-css-selector')` or `click_button 'Foobar'`. To start, you could have a look at some existing feature specs like spec/features/person/people_controller_spec.rb.
Note: To run feature specs locally, you need to pass an additional argument `--tag type:feature`, as documented in the [development repository's README](https://github.com/hitobito/development/#run-desired-tests).

PR #2051, opened 2023-03-26
Imporved invitations

nchiapol:


nchiapol:
I agree that the separat dropdown is a positive thing from a code perspective - but I think from a UI perspective it is wrong. The interface has way too many buttons already - and as a user it feels more natural, when this is all in the same dropdown: 
  1. I want to add people to the event: click
  2. Do I want the participants to confirm or should they be added directly: choose.

But of course I can live with two separat buttons. Thanks for merging anyway.

carlobeltrame:
Please correct me if I'm wrong, but I thought so far there was only one dropdown button visible anywhere, and this dropdown button always either allows to create participations or allows to create invitations, never both. I haven't changed anything visually about this status (except for the label on the button, admittedly). So far, you have to go to the "Teilnehmende" tab to add a participant, or go to the "Einladungen" tab to invite someone. There is no view where both options are offered at the same time (maybe there will be after #2053, I haven't had the time to look at that yet).

What I have changed is only how the dropdown button on the invitations tab works internally. It was intended to be a refactoring, i.e. a code change which makes understanding and working with the code easier, but does not have any noticeable effect for users.

nchiapol:
Thanks for clarifying. Yes I somehow assumed you talked about the button from #2053. Sorry for the confusion. I should look closer at code next time...

PR #2071, opened 2023-04-05
Moved salutation to core and added configurable custom gender (off by default)

lukas-buergi:
This PR goes along with a PR in hitobito_generic which demonstrates the use of the new features in core. Both PRs have one commit for the salutation stuff and one for the gender stuff (so you can take it apart if you want, although the gender stuff won't work without the salutation stuff).

Before: There was already some salutation code in core, but it was incomplete. The gender field was ternary female/male/unknown.

After:

- Salutation (including title) works in core and is a bit more polished than some of the solutions I've seen implemented in wagons.
  - There is a new feature where you can put an arbitrary string in the salutation field using code (not in the UI) and this string is used as is for the entire salutation. When editing the person in the UI, you can keep this custom salutation or switch to one of the configured standard ones. It's intended to make it easier to import datasets with irregular salutations.
  - In the person view the salutation is displayed correctly not as the unformatted template string as before.
  - The salutation is added to invoice exports. Usually we send a letter along with the invoice which needs a salutation.
  - Salutations for households didn't work correctly before, now they do.
  - Made the person edit interface clearer so that the default salutation isn't just a blank field
  - The migrations for this are conditional, but because wagon migrations run after core migrations, all wagons which already have the field still need to be changed to make their migration also conditional in the same way as the migration in the core.
- Gender for salutations: I tried to come up with a pragmatic solution that works for everyone so we wouldn't need to maintain something separate.
  - Added new gender_custom field with choices defined in translation file. This field is disabled by default. In hitobito_generic I enable it for demonstration purposes and call it "Anredengeschlecht" with options male, female, neutral salutation, didn't want to answer question, unknown/blank.
  - If this gender_custom field is set to a non-blank value, it has precedence over the gender field for the salutation. It's not used otherwise.
  - Anything other than male/female gets a neutral salutation, i.e. the same as unknown gender.
  - The old gender field can be disabled with a FeatureGate/config option.
  - To migrate, it's possible to enable the new gender field and have both for a while and eventually maybe disable the old gender field. Or to switch immediately new gender field on, old field off (no data loss). Or to just start with the new one.
  - I'm sure there is a bunch of code that still depends on the old gender field, so if someone wants to disable the old field and needs the functionality there is still work to do. It won't break because the field is just removed from the UI, but it can't be updated anymore from the UI when it's disabled, so...

Tests: Rubocop passes, but a whole bunch of tests fail and I don't see any relation to my changes. If you want to accept the PR and would like me to investigate, just ping me.

### DIFF
--- a/config/settings.yml
+++ b/config/settings.yml
@@ -142,6 +142,10 @@ people:
   default_sort: name
   abos: true
   totp_drift: 15
+  gender_custom:
+    enabled: false
+  gender_binary:
+    enabled: true
 
 event:
   attachments:
--- a/app/controllers/people_controller.rb
+++ b/app/controllers/people_controller.rb
@@ -16,8 +16,8 @@ class PeopleController < CrudController
   self.remember_params += [:name, :range, :filters, :filter_id]
 
   self.permitted_attrs = [:first_name, :last_name, :company_name, :nickname, :company,
-                          :gender, :birthday, :title, :salutation, :additional_information,
-                          :picture, :remove_picture] +
+                          :gender, :gender_custom, :birthday, :title, :salutation,
+                          :additional_information, :picture, :remove_picture] +
                           Contactable::ACCESSIBLE_ATTRS +
                           [family_members_attributes: [:id, :kind, :other_id, :_destroy]] +
                           [household_people_ids: []] +
--- a/app/controllers/people_controller.rb
+++ b/app/controllers/people_controller.rb
@@ -16,7 +16,8 @@ class PeopleController < CrudController
   self.remember_params += [:name, :range, :filters, :filter_id]
 
   self.permitted_attrs = [:first_name, :last_name, :company_name, :nickname, :company,
-                          :gender, :birthday, :additional_information, :picture, :remove_picture] +
+                          :gender, :birthday, :title, :salutation, :additional_information,
+                          :picture, :remove_picture] +
                           Contactable::ACCESSIBLE_ATTRS +
                           [family_members_attributes: [:id, :kind, :other_id, :_destroy]] +
                           [household_people_ids: []] +
--- a/app/domain/export/tabular/invoices/list.rb
+++ b/app/domain/export/tabular/invoices/list.rb
@@ -13,7 +13,7 @@ module Export::Tabular::Invoices
                         cost vat total amount_paid).freeze
 
 
-    CUSTOM_METHODS = %w(cost_centers accounts payments)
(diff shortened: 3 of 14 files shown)


lukas-buergi:
I see that now in the Rails Lint and Test report there is a list of specs that fail because of my changes, somehow those where hidden among lots and lots of others when I tested locally. Most of them seem to just be specs that need to be adjusted slightly, maybe I forgot a FeatureGate test in some cases, I'll have a look soon.

kronn:
I just took a look at this and am wondering: will the title-field break every wagon? I've seen that the generic-wagon needed some changes. This is not a complete blocker, just something I want to be prepared for.

Also, we tend favour `unless` over `if not`. Other details like very short variables (`r`) and missing spaces around `=` are details that we really should have proper linting for. So, those are mostly on me. Could you change those nonetheless?

I would expect for `column_exists?` to be available directly in migrations. If so, the check could be done like this `add_column(...) unless column_exists?(...)`, which reads IMHO better than the three-line incantation.

lukas-buergi:
Nice that you are looking into it, thanks. I can change the small things you mention.

It won't break all wagons, just those adding a title field without checking if it's already there. And it won't break existing deployments, it just happens when the migration in the wagon which adds the field runs. This is because the migrations in the wagons always run after the core migrations, as far as I can tell, so even though the core migration is conditional, because it's the one which runs first, the unconditional wagon migration runs after that and fails. Seems to me like it's easier to do a one-line fix in the wagons than to try to solve this here.

config/settings.yml
```diff
@@ -142,6 +142,10 @@ people:
   default_sort: name
   abos: true
   totp_drift: 15
+  gender_custom:
+    enabled: false
+  gender_binary:
+    enabled: true
```
kronn, line 148: This means that it is possible to have both binary and custom gender-selectors, right? Is that intended? 

It feels like a workaround for the `FeatureGate` missing a proper `else`. If that is the case, I can look into that.

kronn (changes requested):
Could you rebase this onto the current master? The core seems to have changed while I left this hanging (sorry...)

kronn:
One use of the current gender-field is for statistics. Some of the recipients of the statistics only support the distinction between male and female. So it may be that we have gender-buckets (m/f/d/...) and the custom gender used for salutations. This raises the question if the custom gender should be called custom salutation instead. 

I have however not totally wrapped my head around the solution and its edge-cases. 

Maybe a set of specs will help to illustrate?

PR #2086, opened 2023-04-20
Enable dynamic invoice items for letter with invoice

TheWalkingLeek:
Fixes: #2039

### DIFF
--- a/app/controllers/messages_controller.rb
+++ b/app/controllers/messages_controller.rb
@@ -18,14 +18,13 @@ class MessagesController < CrudController
                                     invoice_attributes: {
                                       invoice_items_attributes: [
                                         :name,
-                                        :heading,
                                         :description,
+                                        :cost_center,
+                                        :account,
                                         :unit_cost,
                                         :vat_rate,
                                         :count,
-                                        :cost_center,
-                                        :account,
-                                        :variable_donation,
+                                        :type,
                                         :_destroy
                                       ]
                                     }].freeze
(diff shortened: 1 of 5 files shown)


app/controllers/messages_controller.rb
```diff
@@ -84,7 +97,17 @@ def preview?
   def permitted_params
     p = model_params.dup
     p.delete(:type)
-    p.permit(permitted_attrs)
+    p = p.permit(permitted_attrs)
+
+    return p unless entry.invoice? && p.dig(:invoice_attributes, :invoice_items_attributes)
```
mtnstar, line 102: maybe extract in a getter? 

```suggestion
    return p unless invoice_with_attributes?
```

app/helpers/dropdown/letter_with_invoice_new.rb
```diff
@@ -0,0 +1,62 @@
...
+    end
+
+    private
+
+    def label
+      @label || I18n.t('crud.new.title', model: Message::LetterWithInvoice.model_name.human)
+    end
+
+    def single_button
+      return template.action_button(label, @path, @icon, disabled: @disabled_msg) if @disabled_msg
```
mtnstar, line 34: I think the disabled message feature for this button might not be implemented.

image

I made the invoice_config of one layer invalid and was expecting the button to be disabled. Am I getting it wrong or is this something we have to take care of?

mtnstar (changes requested):
so far so good, only check the disabled_msg feature :partying_face: 
now reviewing also mitte wagon exentions

app/helpers/dropdown/letter_with_invoice_new.rb
mtnstar: maybe add also a spec that checks if button is disabled if the invoice_config is invalid

PR #2095, opened 2023-04-29
fix: enlarged profil image for mobile devices

bergerar:
Limit width of enlarged profile image to be the maximum of the screen width.

Relates to #2044


carlobeltrame (approved):
Merci!

PR #2096, opened 2023-05-01
Enable subscriptions export to export table displays

TheWalkingLeek:
Context: https://help.puzzle.ch/#ticket/zoom/6146

https://github.com/hitobito/hitobito_pbs/pull/276

Even though in the views present, the subscriptions export never (AFAIK) exported table displays


carlobeltrame (approved):
Very strange... But LGTM. A spec would be nice, but if we don't have much time left to work on this, it's fine without.

PR #2102, opened 2023-05-08
Remove migration code which breaks due to missing attribute

TheWalkingLeek:
For some reason, these migrations break due to find_each trying to fetch the column groups.creator_id that does not exist at this point. 
Since these are old data migrations that are applied everywhere and thus not relevant anymore, we chose to just remove these.


kronn (approved):
Schönes aufräumen

PR #2132, opened 2023-06-12
Do not implement logic on class level for permissions

TheWalkingLeek:
This has been a pain because when the Ability constant gets referenced before any course types have been defined (e.g in the wagon.rb of a wagon) the permissions just did not get registered. This happened because when the Ability constant was loaded, the VariousAbility constant was aswell. Thus already evaluating course types before they were loaded

Fixes: https://github.com/hitobito/hitobito_jubla/actions/runs/5239299435/jobs/9458999780
Fixes: #2113


kronn (approved):
A similar issue bit me elsewhere (with mails, I think?) a while back.

Your conclusion is the same a mine, so logically we must both be stupid or genius. I naturally opt for the later.

PR #2135, opened 2023-06-13
Use ruby sort for amount

TheWalkingLeek:
This was a problem since either ActiveRecord or MySQL made an error when sorting the amount which is saved in scientific E notation (0.1e2 = 10)

Refs: https://help.puzzle.ch/#ticket/zoom/6293


kronn (approved):
The need to do this makes me sad, the code/test-ratio makes me happy.

PR #2138, opened 2023-06-19
Feature/2136 last active role as attribute

### LINKED ISSUE #2136: PEOPLE: Letzte aktive Rolle auf Person speichern
Das ermitteln von "Personen ohne Rollen" ist technisch jedes Mal ein ziemlicher Aufwand. Insbesondere da es ein Zustand einer Person ist, der nur selten ändert.

Daher ist das Ziel nun, ein separates Feld auf der Person zu hinerlegen welches die ID zur letzen soft-gelöschten Rolle hinterlegt hat.
Das Feld ist nur befüllt wenn die Person sonst keine aktive Rolle hat.

# Tech-Spec

- Umsetzung im Core

# ToDo

- [ ] Migration erstellen: `last_active_role_id`
- [ ] RecurringJob anlegen welcher täglich(?) läuft
  - [ ] befüllt `last_active_role_id` wenn nötig
  - [ ] entfernt `last_active_role_id` wenn nötig
- [ ] `Group::DeletedPeople` anpassen sodass es mit `last_active_role_id` arbeitet
- [ ] Evtl. gibt es noch andere Orte an denen die Personen ohne Rollen gesammelt werden, diese auch anpassen
- [ ] Wagons überprüfen
- [ ] Sicherstellen dass alle Specs laufen
- [ ] Specs schreiben
- [ ] Volltextsuche Performance testen
- [ ] Mit angemessener Rolle "durchklicken"

nilsrauch:
Closes: #2136 

~~Job hatte lokal mit einem Mitte Prod dump ca. 40 Sekunden~~ job wurde entfernt, wir haben eine lösung ohne gefunden


mtnstar:
as just discussed, I think we can get rid of the `AssignLastActiveRoleJob` if we concern the `deleted_at` date of the last active role in the SQL query. 

so if `last_active_role_id` is set on a person, we're checking if the `deleted_at` property of this last active role is either today or in the past. 

so we're going to need a callback in the Role model which sets the correct `last_active_role_id` based on all Roles' `deleted_at` attribute.

set_last_active_role callback:

1. do nothing unless there was a change to `deleted_at`
2. if `deleted_at` was changed, so check if all the other present roles have `deleted_at` set
3. if ALL person's roles do have  `deleted_at` set, get the one which is the most in the future
4. set `last_active_role_id` to the role id of the selected role in step 3.

app/models/role.rb
```diff
@@ -242,4 +244,15 @@ def prevent_changes
 
     raise ActiveRecord::ReadOnlyRecord unless new_record? || only_archival
   end
+
+  def set_last_active_role
+    still_has_active_role = person.roles.with_deleted.where(roles: { deleted_at: nil }).present?
```
mtnstar, line 249: do we need to use the `with_deleted` scope here? 
doesn't the default scope for roles already return only those where `deleted_at: nil` ?

db/migrate/20230615084630_add_last_active_role_to_people.rb
```diff
@@ -0,0 +1,38 @@
...
+#  https://github.com/hitobito/hitobito.
+
+class AddLastActiveRoleToPeople < ActiveRecord::Migration[6.1]
+  def up
+    add_column :people, :last_active_role_id, :integer, null: true
+
+    say_with_time('assigning last_active_role') do
+      people_without_active_roles.find_each do |person|
+        person.update_column(:last_active_role_id,
+                             person.roles.only_deleted.order(deleted_at: :desc).first.id)
```
mtnstar, line 15:
```suggestion
                             person.roles.inactive.order(deleted_at: :desc).first.id)
```

db/migrate/20230615084630_add_last_active_role_to_people.rb
```diff
@@ -0,0 +1,38 @@
+# frozen_string_literal: true
+
+#  Copyright (c) 2023, Hitobito AG. This file is part of
+#  hitobito and licensed under the Affero General Public License version 3
+#  or later. See the COPYING file at the top-level directory or at
+#  https://github.com/hitobito/hitobito.
+
```
mtnstar, line 7: maybe make use of newly introduced role scopes in this migration

spec/migrations/add_last_active_role_to_people_spec.rb
```diff
@@ -0,0 +1,69 @@
+# frozen_string_literal: true
```
mtnstar, line 1: nice ...we even have a spec for the migration, love it :partying_face:

mtnstar (changes requested):
it's very nice that we invest some time into this feature since I think we're going to introduce also roles that are active in the future when created. 

some naming stuff I also was thinking about ...

would be great to chat a bit about this important feature/fix

app/models/role.rb
TheWalkingLeek: Nope, returns `deleted_at IS NULL OR deleted_at >= :now`

db/migrate/20230615084630_add_last_active_role_to_people.rb
kronn: thanks, @carlobeltrame. Scopes change (with this PR, for example) and the App relies on the migrations which rely on the Database, not the other way around.

spec/migrations/add_last_active_role_to_people_spec.rb
```diff
@@ -0,0 +1,69 @@
...
+
+  let(:migration) { described_class.new.tap { |m| m.verbose = false } }
+
+  after do
+    Person.delete_all
+  end
+
+  context '#up' do
+    let!(:people_without_active_roles) do
+      10.times.collect do
```
amaierhofer, line 26: Here we are testing the same thing with 10 people (which does not add value imho but slows down tests)

spec/migrations/add_last_active_role_to_people_spec.rb
```diff
@@ -0,0 +1,69 @@
...
+        Fabricate(Group::BottomLayer::LocalGuide.sti_name.to_sym,
+                  created_at: 1.month.ago, deleted_at: 1.week.ago,
+                  group: groups(:bottom_layer_one),
+                  person: person)
+        person
+      end
+    end
+
+    let!(:people_with_active_roles) do
+      10.times.collect do
```
amaierhofer, line 40: same here

amaierhofer (commented):
I still see some potential and I believe the case of having a future `deleted_at` date on the role, which you can set via the UI is not handled yet. This probably requires a cronjob checking on those roles regularly

amaierhofer (changes requested):
- [ ] spec out role callbacks
- [ ] implement job that handles roles with `delete_at` date in the future (see also #1748, these cases should also be handled by such a job)

carlobeltrame:
> I still see some potential and I believe the case of having a future `deleted_at` date on the role, which you can set via the UI is not handled yet. This probably requires a cronjob checking on those roles regularly

Absolutely, thanks for noticing! See also #1748 where I tried to collect similar existing bugs. Ideally, such a "after-destroy-cleanup" cronjob would run all the (potentially wagon-specific) logic which should run on the deletion date of a role.

PR #2152, opened 2023-06-29
Update gems to make hitobito run with ruby 3.0

mtnstar:
Refs: #2128 
noch offene Tasks:

- [x] Letzte specs noch fixen
- [ ] upgrade auf ruby 3.2


kronn:
included in #2213

PR #2208, opened 2023-09-30
removes "deleted subgroups"- tab for unnecessary roles

philobr:
The tab for deleted subgroups was visible for everyone. Now only the following permissions can access it: 
- layer and below full
- layer full
- group and below full
- group full


TheWalkingLeek:
Since the branch needed a rebase and some spec fixes, I continued this in #2269 which is now merged, thanks for your contribution :)

PR #2344, opened 2024-01-06
rszr

kronn:
In order to reduce the memory footprint, image size, attack surface and overall complexity, this PR removes ImageMagick (used through `mini_magick`) and adds libimlib2 (used through `rszr`).

I transferred all processing instructions from the previous carrierwave-uploaders. Also, this adds auto-rotation to all uploads. Therefore, this offers more features than before were available. The loss of the auto-rotation was a regression of the carrierwave->activestorage migration.

See #1658, needs #2343

### DIFF
--- a/.github/workflows/sbom.yml
+++ b/.github/workflows/sbom.yml
@@ -71,6 +71,7 @@ jobs:
         run: |
           sudo apt-get -qq update
           sudo apt-get install sphinxsearch
+          sudo apt-get install libimlib2 libimlib2-dev
           echo "ruby version: $(ruby -v)"
           echo "node version: $(node -v)"
           echo "yarn version: $(yarn -v)"
--- a/.github/workflows/security-scanners.yml
+++ b/.github/workflows/security-scanners.yml
@@ -12,6 +12,11 @@ jobs:
       - name: Checkout code
         uses: actions/checkout@v4
 
+      - name: Install dependencies
+        run: |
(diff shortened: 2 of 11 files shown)


kronn:
Currently, the base it set to the cleanup-carrierwave branch for easy comparison. Depending on the order which these PRs are merged, a change of ~pace~ base might be needed.

kronn:
Uploads and resizing work as before, the size of the generated Dockerimage is however vastly reduced. From a local build of the generic/demo-composition repo:

```
❯ docker images --format "{{.Repository}}:{{.Tag}} -> {{.Size}}" hitobito/demo/rails | sort
hitobito/demo/rails:latest -> 1.13GB
hitobito/demo/rails:rszr -> 886MB
```

The only difference was in the hitobito-state: "current master" (835c46bdf) versus "rszr-branch" (d480b2bd7)

TheWalkingLeek (approved):
Schaut gut aus, ich teste das noch auf paar Integrationen und merge dann :)

TheWalkingLeek:
Hatte das mal auf der Pro Natura Integration deployt gehabt und anschliessend gingen die Bilduploads bei den Personen nicht mehr. Habs nur bei den Personen getestet iirc
Fehlermeldung war, dass das Bild nicht dem verlangten Format entspricht

kronn:
We will switch to vips with the next rails-upgrade. therefore I'll close this

PR #2346, opened 2024-01-08
Delegate FutureRole#to_s to new role

daniel-illi:
fixes hitobito/hitobito_sac_cas#174


daniel-illi:
let's not modify the core

PR #2354, opened 2024-01-10
Speed up people#index by removing 3n+1 queries

amaierhofer:
Seems to improve people#index page by about 30-50%

- include picture_attachment in scope
- filter preloaded roles in memory


kronn (changes requested):
I'd mostly like to see the spec improved. This way, it may have helped you developing it. It will not help me understand the problem if the spec fails.

PR #2386, opened 2024-01-26
Refactor delayed_jobs tasks, allow adding jobs in wagons

daniel-illi:


app/jobs/job_manager.rb
```diff
@@ -0,0 +1,81 @@
...
+      job_class.new.schedule
+    end
+  end
+
+  def clear
+    Delayed::Job.delete_all
+  end
+
+  def check
+    scheduled, missing = jobs.partition { _1.new.scheduled? }
```
kronn, line 22: I'll be happy once we have Ruby 3.4 and can write `it.new.scheduled?` here instead.

For now, it is inconsistent with line 12 above. How about:

```suggestion
    scheduled, missing = jobs.partition do |job_class|
      job_class.new.scheduled?
    end
```

app/jobs/job_manager.rb
```diff
@@ -0,0 +1,81 @@
...
+      DownloadCleanerJob,
+      Payments::EbicsImportJob,
+      People::DuplicateLocatorJob,
+      People::CleanupJob,
+      People::CreateRolesJob,
+      People::DestroyRolesJob,
+      ReoccuringMailchimpSynchronizationJob,
+      SessionsCleanerJob,
+      WorkerHeartbeatCheckJob,
+      Event::ParticipationCleanupJob
```
kronn, line 78: How about sorting them again, while we're at it?

```suggestion
      DownloadCleanerJob,
      Event::ParticipationCleanupJob,
      Payments::EbicsImportJob,
      People::DuplicateLocatorJob,
      People::CleanupJob,
      People::CreateRolesJob,
      People::DestroyRolesJob,
      ReoccuringMailchimpSynchronizationJob,
      SessionsCleanerJob,
      WorkerHeartbeatCheckJob
```

kronn (changes requested):
I like it, but I have minor nit-picking unrelated to the main addition.

PR #2398, opened 2024-02-01
Repair OAuth::Application images

kronn:
The correct default image was not referenced. 

This was an oversight of #1658 

While at it, the default has been corrected and the invocation changed to a more stable one.

See https://help.puzzle.ch/#ticket/zoom/7277


kronn:
Yesterday? no.
This night? yes. 

We update the integration-instances every weekday around 3 in the night. So, by now, it should be there.

PR #2400, opened 2024-02-02
Feature/graphiti tooling improvements

### LINKED ISSUE #2399: Improve graphiti tooling
The current spec for checking the graphiti schema disrupts fixture setup in wagon specs.
Besides it adds a penalty to each spec run as the implicit spec can not be skipped easily.

* Use a regular, explicit spec to check if the schema file exists and is up to date on CI
  - shared example in core
  - spec in wagon that uses the shared example (add also to wagon templates)
* Add rake tasks to manage the schema file

daniel-illi:
fixes #2399


lib/tasks/graphiti.rake
```diff
@@ -0,0 +1,96 @@
...
+    desc 'Check if the schema file exists'
+    task file_exists: 'graphiti:schema:setup' do
+      abort red(<<~MSG) unless File.exist?(Graphiti.config.schema_path)
+        Schema file not found: #{pretty_schema_path}
+
+        Run `bundle exec rake graphiti:schema:generate` and commit the file to the git repository.
+
+      MSG
+
+      puts "\e[32mSchema file exists\e[0m"
```
kronn, line 32:
```suggestion
      puts green('Schema file exists')
```

lib/tasks/graphiti.rake
```diff
@@ -0,0 +1,96 @@
...
+          Schema file is outdated: #{file}
+
+          Run `bundle exec rake graphiti:schema:generate` and commit the file to the git repository.
+        MSG
+      end
+
+      puts green 'Schema file is up to date'
+    end
+
+    desc 'Check if the schema is backwards-compatible'
```
kronn, line 72:
```suggestion
    desc 'Generate or update the schema file'
```

kronn (changes requested):
I found two small improvements, but in general, I like it.

PR #2406, opened 2024-02-07
Setup TOTP for local dev-admins

kronn:
When working in a wagon that has enforced 2FA, TOTP needs to be setup upon first login. If you create a new local admin, it most certainly has a role that needs this.

This PR set the secrets in a way that is repeatable and therefore reduces the friction when supporting organizations using such secure wagon setups. The same 2FA-setup can then be used again for development. It is not intended for production or integration setups, it should be (as before) be purely local admin.

### DIFF
--- a/lib/tasks/dev.rake
+++ b/lib/tasks/dev.rake
@@ -47,6 +47,8 @@ namespace :dev do
     desc 'Create a local user with admin-permissions'
     task admin: :environment do
       abort('This is for development purposes only.') unless Rails.env.development?
+      abort('This needs at least one wagon to work') if Wagons.all.blank?
+      abort('This needs a group-structure to work') if Group.subclasses.blank?
 
       username = 'tester@example.net'
       password = 'hitobito is the best software to manage people in complex group hierachies'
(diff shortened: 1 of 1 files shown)


lib/tasks/dev.rake
```diff
@@ -91,6 +91,42 @@ namespace :dev do
...
+
+        otp = People::OneTimePassword.new(me.two_fa_secret, person: me)
+        otp.provisioning_qr_code.save(qr_code.to_s)
+
+        case ENV.fetch('TERM', nil)
+        when 'xterm-kitty'
+          puts 'This is the QR-Code for the TOTP/2FA-Setup'
+          system("kitty +kitten icat #{qr_code}")
+        else
+          puts "The QR-Code for TOTP/2FA-Setup is located a #{qr_code}"
```
KramNamez, line 115: "located at" <- Typo fix.

KramNamez:
Just to be sure: This needs to be activated explicitly, in a non-default way, where it is clear that this is for local development only?

Then this is fine. 
If it is easy for someone to accidentally turn this on, it is not.

kronn:
> Just to be sure: This needs to be activated explicitly, in a non-default way, where it is clear that this is for local development only?
> 
> Then this is fine. If it is easy for someone to accidentally turn this on, it is not.

The command to activate this is called "rake dev:local:admin". It is possible to run this on a production cluster, but to do so you would need to already have control over arbitrary commands and could then do all of this one by one.

To complicate this shortcut for hackers, I added a guard that only executes this in development-mode but aborts otherwise.

PR #2424, opened 2024-02-21
Sort invoices by date of last payment

kronn:
This adds a new column "last payment at" to the invoices-tables and allows sorting by it.
It further adds the sorting for "amount paid", which has been missing (it was displayed, but did not work).
The invoice-area is now wrapped in a group-sheet to reduce the disruptive design of it. In future changes to this area, a more common navigation (tabs) might be added.

This also contains a bug fix for when invalid dates are entered into the date-range fields.

Fixes hitobito/hitobito_sww#147

### DIFF
--- a/app/models/invoice.rb
+++ b/app/models/invoice.rb
@@ -149,11 +149,16 @@ class Invoice < ActiveRecord::Base
       'last_payments.received_at'
     end
 
+    def order_by_amount_paid_statement
+      'last_payments.amount_paid'
+    end
+
     def last_payments_information
       <<~SQL.squish
         LEFT OUTER JOIN (
           SELECT invoice_id,
-                 MAX(received_at) AS received_at
+                 MAX(received_at) AS received_at,
+                 SUM(amount) AS amount_paid
           FROM payments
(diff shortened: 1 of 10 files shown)


app/models/invoice.rb
```diff
@@ -113,10 +113,8 @@ def draft_or_issued_in(year)
     end
 
     def draft_or_issued(from:, to:)
-      from ||= Time.zone.today.beginning_of_year
-      to ||= Time.zone.today.end_of_year
-      from = Date.parse(from) if from.is_a? String
-      to = Date.parse(to) if to.is_a? String
+      from = Date.parse(from.to_s) rescue Time.zone.today.beginning_of_year # rubocop:disable Style/RescueModifier
+      to = Date.parse(to.to_s) rescue Time.zone.today.end_of_year # rubocop:disable Style/RescueModifier
```
TheWalkingLeek, line 117: Kannte ich nicht, in dem Fall ziemlich nice

TheWalkingLeek (approved):
Habs noch manuell durchgeklickt und funktionierte wie erwartet :)

PR #2494, opened 2024-03-13
Silence low-contrast warnings by lowering our standards

kronn:
The default contrast-ratio of 4.5:1 triggers warnings. I noticed it in the skv-wagon, I did not research any further.

Discussion welcome


carlobeltrame (changes requested):
Since hitobito is a software for a very broad range of users, I think we should not make assumptions about the absence of visual impairments in our userbase. I would much rather fix the contrast issues themselves.

PR #2498, opened 2024-03-14
Handle and avoid breaking changes in cdxgen

carlobeltrame:
https://github.com/CycloneDX/cdxgen/pull/837 removed the XML format. Fortunately, our CLI SBOM merge tool can also handle JSON. Hopefully, DependencyTrack can also handle it.

To avoid future breakage, the major version of the package is pinned.

### DIFF
--- a/.github/workflows/sbom.yml
+++ b/.github/workflows/sbom.yml
@@ -149,19 +149,19 @@ jobs:
         env:
           SCAN_DEBUG_MODE: debug
         run: |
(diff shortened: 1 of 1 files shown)


kronn (approved):
Maybe ask @KramNamez about insights on Dependency Track.

Otherwise, this looks good to me.

KramNamez:
Should work fine - I haven't tested switching formats, but Deptrack definitely accepts JSON input too.

PR #2499, opened 2024-03-14
Toggle the correct css hidden class after turbo changes

TheWalkingLeek:
Fixes: #2496


carlobeltrame (approved):
Tested, works, thanks!

PR #2508, opened 2024-03-18
Allow to display and export additional information

kronn:
See #2421

### DIFF
--- a/app/domain/export/pdf/participation/section.rb
+++ b/app/domain/export/pdf/participation/section.rb
@@ -1,6 +1,6 @@
 # frozen_string_literal: true
 
-#  Copyright (c) 2012-2021, Jungwacht Blauring Schweiz. This file is part of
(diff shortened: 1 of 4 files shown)


carlobeltrame (approved):
Code looks good, and my manual tests were successful. Still trying to get the CI to pass without flaking.

kronn:
@carlobeltrame I can reproduce the error locally (in specs), even on master. Therefore it seems unrelated to my change. I am at a loss how to repair this correctly. Just adapting the specs (even just with different margins, so it does not wrap) seems wrong to me.

carlobeltrame:
@kronn I think it's likely that on some days of the year, the date and/or time take up more space than on other days, and therefore the text wraps. Our approach to testing this PDF is flawed.

Nevermind this specific failing test, but I'd like to know that the test suite otherwise passes, especially since you did not adjust any existing specs. Can you run the feature specs locally and confirm that everything still works? Then I'd be fine to merge this as is.

PR #2517, opened 2024-03-20
RSpec binstub

kronn:
This does:

- cleanup of some `wagon`-commands
- assume that we mostly test the core from the core
- assume better integration with tooling through a dedicated bin-stub

I tested (without using spring)

- rails db:drop db:create # from core
- wagon reset-test-database # from core
- wagon reset-test-database # from wagon
- wagon test-prepare # from core
- wagon test-prepare # from wagon
- rspec --fail-fast --seed 58558 # from core
- rspec # from wagon
- :TestFile # from core, in neovim with vim-test and dispatch


carlobeltrame (approved):
Habs kurz mit dem Docker Setup getestet, scheint gleich wie vorher zu funktionieren

PR #2529, opened 2024-03-26
Hint at disallowed content

kronn:
See #2421 and cevi/hitobito_cevi#183

If permissions exist to views the content, it will be displayed. Ohterwise a hint "fehlende Berechtigung" is shown. This should clarify the difference between "no info" and "no permission".


TheWalkingLeek (approved):
Suberi Sach. Manuelle tests liefen auch durch :+1:

PR #2535, opened 2024-03-27
Disable turbo on (person/csv-import) upload-form

kronn:
The upload-form was not advancing to the next step. without turbo, it worked again fine. I did not do a full research on this.


daniel-illi (approved):
LGTM :+1:

PR #2566, opened 2024-04-12
Add event number filter to json:api event resource

simfeld:
cc @gexclaude

### DIFF
--- a/spec/resources/events/reads_spec.rb
+++ b/spec/resources/events/reads_spec.rb
@@ -124,6 +124,23 @@ describe EventResource, type: :resource do
       end
     end
 
+    describe 'by number' do
+      let!(:one) { Fabricate(:event, number: 'COURSE-123') }
+      let!(:two) { Fabricate(:event, number: 'COURSE-456') }
+
+      it 'returns only events with equal number' do
+        params[:filter] = { number: course.number }
+        render
+        expect(jsonapi_data).to have(1).items
+      end
+
+      it 'returns only events wih number having a specific prefix' do
+        params[:filter] = { number: { prefix: 'COURSE-' } }
+        render
+        expect(jsonapi_data).to have(2).items
+      end
+    end
+
     describe 'by before_or_on' do
       it 'filters using Event::before_or_on' do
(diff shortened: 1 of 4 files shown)


spec/resources/events/reads_spec.rb
```diff
@@ -124,6 +124,23 @@
...
+      let!(:one) { Fabricate(:event, number: 'COURSE-123') }
+      let!(:two) { Fabricate(:event, number: 'COURSE-456') }
+
+      it 'returns only events with equal number' do
+        params[:filter] = { number: course.number }
+        render
+        expect(jsonapi_data).to have(1).items
+      end
+
+      it 'returns only events wih number having a specific prefix' do
```
simfeld, line 137:
```suggestion
      it 'returns only events with number having a specific prefix' do
```

spec/support/resource_spec_helper.rb
```diff
@@ -10,7 +10,7 @@ module ResourceSpecHelper
 
   included do
     let(:ability) { Ability.new(person) }
-    let(:person) { people(:top_leader) }
+    let(:person) { Fabricate(:person) }
```
simfeld, line 13: See https://github.com/simfeld/hitobito_pbs/actions/runs/8669188064/job/23775501581
```
Failure/Error: raise StandardError, "No fixture named '#{f_name}' found for fixture set '#{fs_name}'"

     StandardError:
       No fixture named 'top_leader' found for fixture set 'people'
```

kronn (approved):

kronn:
I tried to rebase this PR so it can be merged. Unfortunately, the number is usually not an attribute on events. That's why it doesn't work that way. Either this would have to be solved directly in the wagon or only for courses. But I could not do it quickly.

I generally think filtering by number is a good idea. Could you update this PR accordingly?

kronn:
Nevermind, I found the problem I had, will resolve it and merge it.

Events normally do not have a number, so the attribute currently is only present on courses. When moving the change and the specs to the course-part, it works like it should.

simfeld:
@kronn thanks for finding the proper fix 🚀

PR #2568, opened 2024-04-15
Add optional locale argument to Country#label, refs hitobito/hitobito_sac_cas#427

daniel-illi:
fixes hitobito/hitobito_sac_cas#427


daniel-illi:
localized country label is not required anymore

PR #2599, opened 2024-05-06
Try to use a common branch for all repos

kronn:
When testing the wagon, it tries to find a matching branch in the core and youth-wagon.
If the feature-branch cannot be found, it should default to the value from the input, which in turn defaults to `master`.

This allows to have synced changes in the wagon and the core, just by naming the feature-branch the same.

I only tested the output of the `git ls-remote | cut | grep`-command, manually outside the action. A full integration-test is of course possible, but IMHO not worth the effort. If needed, I would revert and redo this.

### DIFF
--- a/.github/workflows/wagon-tests.yml
+++ b/.github/workflows/wagon-tests.yml
@@ -14,11 +14,11 @@ on:
       core_ref:
         description: Use a specific version of the core for the workflow run. Defaults to master.
         type: string
-        default: 'master'
+        default: ''
       wagon_dependency_ref:
         description: Use a specific version of the wagon dependency for the workflow run. Defaults to master.
         type: string
-        default: 'master'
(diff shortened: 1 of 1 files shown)


carlobeltrame (commented):
I'd prefer if a manual input always overrides any automatic fallback. Can you check if the manual input differs from master and in that case use the manual input?

kronn:
> I'd prefer if a manual input always overrides any automatic fallback. Can you check if the manual input differs from master and in that case use the manual input?

I removed the defaults that were scattered at various places and handled it completely in the step I added. Now, every provided value (even master) has precedence over the default empty string. Otherwise, you could not force the tests to run against master if a matching branch was present.

TheWalkingLeek (approved):

PR #2646, opened 2024-06-04
Use hpricot from GitHub which builds on newer systems

rnestler:
Fixes a build error during bundle install:

    Installing hpricot 0.8.6 with native extensions
    Gem::Ext::BuildError: ERROR: Failed to build gem native extension.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -111,6 +111,8 @@ group :development, :test do
   gem 'binding_of_caller'
   gem 'codez-tarantula', require: 'tarantula-rails3'
   gem 'graphiti_spec_helpers'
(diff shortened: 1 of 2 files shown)


kronn (approved):
We could think about removing or updating tarantula as it is not being used right now. But that is a different discussion for the whole team.

Your change looks good :-)

rnestler:
> We could think about removing or updating tarantula as it is not being used right now. But that is a different discussion for the whole team.

@kronn Since it looks quite unmaintained that would probably be better. Should I open a PR removing tarantula instead?

kronn:
Removal would entail some cleanup of rake-tasks and such... Also, I be that to be team-decision.

For now, I would take this PR and ask the team tomorrow. Decision will take until end of the week, I guess.

PR #2649, opened 2024-06-04
Make VerificationQrCode#verify_url public

rnestler:
That way we can easily link the QR image to the target URL.

Refs: https://github.com/hitobito/hitobito_sac_cas/issues/533


mtnstar:
looks pretty sane ... usually we create only one PR per feature/issue and per repository.
so maybe there's some other changes in the core for this feature?
please also add a refs #ticket-nr in the description of the PR ... so it gets linked with the ticket

PR #2652, opened 2024-06-05
Don't run tests twice on pull-requests

rnestler:
When the trigger is on all branches and all pull-requests, the tests will run twice when not originating from a fork. Thus we limit it to only master.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -3,6 +3,7 @@ name: 'Rails Lint and Test'
 on:
   workflow_dispatch:
   push:
+    branches: [ master ]
     paths-ignore:
       - 'doc/**'
       - '**.md'


.github/workflows/tests.yml
```diff
@@ -3,6 +3,7 @@ name: 'Rails Lint and Test'
 on:
   workflow_dispatch:
   push:
+    branches: [ master ]
```
rnestler, line 6: Does it need to run on other stable branches as well? `sac-master` maybe? But it seems it is just a delayed state of `master` anyways.

carlobeltrame:
We previously did this already in #1935 and then reverted it later. I personally see value in both versions of the CI running, and would like to keep them. The only case where I can see a redundancy is the push run *when there is also a pull request open for the same branch*. However that is way harder to check for in github actions.

kronn (approved):

PR #2655, opened 2024-06-05
Validate presence of parent in groups

rnestler:
We noticed this while investigating the seeding failure, which was wrongly fixed in https://github.com/hitobito/hitobito/pull/2648.

Except for the `root` groups every group should have a parent.

### DIFF
--- a/app/models/group.rb
+++ b/app/models/group.rb
@@ -168,7 +168,7 @@ class Group < ActiveRecord::Base # rubocop:disable Metrics/ClassLength
 
   validates_by_schema except: [:logo]
   validates :type, uniqueness: { scope: :parent_id }, if: :static_name
-  validates :parent, presence: true, unless: -> { Group.roots.count.zero? || id == Group.root.id }
+  validates :parent, presence: true, unless: :layer?
   validates :name, presence: true, unless: :static_name
   validates :email, format: Devise.email_regexp, allow_blank: true
   validates :description, length: { allow_nil: true, maximum: (2**16) - 1 }
--- a/app/models/group.rb
+++ b/app/models/group.rb
@@ -168,6 +168,7 @@ class Group < ActiveRecord::Base # rubocop:disable Metrics/ClassLength
 
   validates_by_schema except: [:logo]
   validates :type, uniqueness: { scope: :parent_id }, if: :static_name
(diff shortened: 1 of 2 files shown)


mtnstar (approved):

mtnstar:
@rnestler what comes up my mind ... maybe add a spec for this in the models group spec?

app/models/group.rb
```diff
@@ -168,6 +168,7 @@ class Group < ActiveRecord::Base # rubocop:disable Metrics/ClassLength
 
   validates_by_schema except: [:logo]
   validates :type, uniqueness: { scope: :parent_id }, if: :static_name
+  validates :parent, presence: true, unless: :root?
```
rnestler, line 171: I noticed that this is tautological. `:root?` comes form the awesome_nested_set gem and is defined like this:

https://github.com/collectiveidea/awesome_nested_set/blob/master/lib/awesome_nested_set/model.rb#L129-L131

rnestler: But if there can be only one `root` we can check if `Groop.root` exists.

app/models/group.rb
```diff
@@ -168,6 +168,7 @@ class Group < ActiveRecord::Base # rubocop:disable Metrics/ClassLength
 
   validates_by_schema except: [:logo]
   validates :type, uniqueness: { scope: :parent_id }, if: :static_name
+  validates :parent, presence: true, unless: -> { Group.roots.count.zero? || id == Group.root.id }
```
rnestler, line 171: This should work in the following edge cases I thought off:
 * Creating the first group: `Group.roots.count.zero?` will be true
 * The root group will still be valid when it has been created since `id == Group.root.id` has been created.
 * Further groups must have a `parent` present.

app/models/group.rb
rnestler: OK I'm not sure anymore if this is the way to go, since now a lot of other specs fail.

rnestler: Changed it again to `unless: layer?`, which seems feasible.

mtnstar:
@rnestler thx for your thoughts about this ... maybe we can have a chat next week. 
this is not urgent, so better taking some time and making it right

PR #2658, opened 2024-06-06
Remove tarantula

kronn:
This PR remove tarantula from the Gemfile, including its custom dependency. Also, it removes the dedicated templates for new wagons, the test-helpers and -runner in this repo. Last but not least, it updates the docs. But you see all that from the diff.

We decided that updating tarantula to work with Rails 6.1 and newer versions would require more effort than we want to invest right now. Also, our test-suite has grown enough to cover many cases that were covered by tarantula before, and will continue to grow when we discover new bugs and their fixes.

The cleanup in the wagon will be done without ceremony after this has been accepted.


doc/development/04_wagons.md
```diff
@@ -78,10 +78,10 @@ If the wagon is not for a specific organisation and does not define a group stru
 * Translations of those models in `config/locales/models.[name].de.yml`
 * seed-data: `rm -rf db/seeds`
 
-In order to have useful Testdata and to use tarantula, adjust the fixtures in the wagon according to the generated organizational structure:
+In order to have useful Testdata, adjust the fixtures in the wagon according to the generated organizational structure:
 
-* Fixtures for people, groups, roles, events, ... (`spec/fixtures`) (Groups can be exported with `rake fixtures:groups`)
-* Adjusting the tarantula tests in the wagon (`test/tarantula/tarantula_test.rb`)
+* Fixtures for people, groups, roles, events, ... (`spec/fixtures`)
+* Groups can manually created and then exported with `rake fixtures:groups` to have a realistic and usable set.
```
rnestler, line 84:
```suggestion
* Groups can be created manually and then exported with `rake fixtures:groups` to have a realistic and usable set.
```

lib/templates/wagon/test/test_helper.rb.tt
```diff
@@ -1,18 +0,0 @@
...
-
-class ActiveSupport::TestCase
-  # Setup all fixtures in test/fixtures/*.(yml|csv) for all tests in alphabetical order.
-  reset_fixture_path File.expand_path('../../spec/fixtures', __FILE__)
-  #
-  # Note: You'll currently still have to declare fixtures explicitly in integration tests
-  # -- they do not yet inherit this setting
-  fixtures :all
-
-end
```
rnestler, line 18: This looks unrelated to tarantula.

rnestler (approved):
LGTM :tada: a minor fix and I'm not sure about one of the test helpers.

lib/templates/wagon/test/test_helper.rb.tt
kronn: Only the tarantula-suite used the test-helper everything else is RSpec.

PR #2675, opened 2024-06-10
Fix charlock_holmes build on newer systems

rnestler:
On my system charlock_holmes fails to build, since it overrides C++ compiler flags.

See https://github.com/brianmario/charlock_holmes/pull/174

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -20,7 +20,7 @@ gem 'bleib', '~> 0.0.10'
 gem 'bootsnap', require: false
 gem 'cancancan', '< 3.2.0'
 gem 'caxlsx'
-gem 'charlock_holmes', '~> 0.7.7'
+gem 'charlock_holmes', github: 'rnestler/charlock_holmes', ref: 'remove-std-override'
 gem 'commonmarker'
 gem 'config'
 gem 'country_select'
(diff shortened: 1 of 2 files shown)


kronn (changes requested):
I somehow want to have your fix, but distributed as a gem. Do you have expectations when it might be merged?

Alternatively, could you create a local workaround to fix your setup?

At the very least, I would like to see a comment which PR and issue forced us to have a git dependency for this.

rnestler:
> Alternatively, could you create a local workaround to fix your setup?

Sadly not yet. I think a local workaround would be possible with the latest master which has https://github.com/brianmario/charlock_holmes/pull/153 merged.

rnestler:
Hah I found one: `gem install charlock_holmes -- --with-cxxflags='#'` this will comment out the custom cxxflags of the package. Feels like a shell injection  :laughing:

rnestler:
Since I found a workaround I guess we can close this.

kronn:
> Hah I found one: `gem install charlock_holmes -- --with-cxxflags='#'` this will comment out the custom cxxflags of the package. Feels like a shell injection  :laughing: 

I guess technically, it is.

I can't help but think "little Bobby Cflags, we call him..." https://xkcd.com/327/ :joy:

PR #2677, opened 2024-06-11
Feature/2673 Sort group names locale aware

rnestler:
image image

Ref: #2673

### DIFF
--- a/app/helpers/sheet/group/nav_left.rb
+++ b/app/helpers/sheet/group/nav_left.rb
@@ -39,7 +39,7 @@ module Sheet
         end
 
         children_of.each_value do |children|
-          children.sort_by!(&:display_name)
+          children.sort_by!(&:sorting_name)
         end
 
         results = []
@@ -141,7 +141,6 @@ module Sheet
 
       def render_sub_layers
         layers = grouped_sub_layers.map do |type, layers|
-          layers.sort_by!(&:display_name)
           content_tag(:li, content_tag(:span, type, class: 'divider')) +
             safe_join(layers.map { |l| render_sub_layer(l.decorate)  })
 
--- a/app/helpers/sheet/group/nav_left.rb
+++ b/app/helpers/sheet/group/nav_left.rb
@@ -31,8 +31,31 @@ module Sheet
 
       private
 
+      def sort_list_by_name(objects)
+        children_of = {}
+        objects.each do |o|
+          children_of[o.parent_id] ||= []
+          children_of[o.parent_id] << o
+        end
+
+        children_of.each_value do |children|
+          children.sort_by! { |c| c.short_name || c.name }
+        end
+
+        results = []
+        recombine_lists(results, children_of, nil)
+        results
+      end
+
+      def recombine_lists(results, children_of, parent_id)
+        children_of[parent_id]&.each do |o|
+          results << o
+          recombine_lists(results, children_of, o.id)
+        end
+      end
+
       def groups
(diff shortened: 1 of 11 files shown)


app/helpers/sheet/group/nav_left.rb
```diff
@@ -31,8 +31,31 @@ def render
...
+        recombine_lists(results, children_of, objects.first.parent_id)
+        results
+      end
+
+      def recombine_lists(results, children_of, parent_id)
+        children_of[parent_id]&.each do |o|
+          results << o
+          recombine_lists(results, children_of, o.id)
+        end
+      end
```
rnestler, line 55: Mostly inspired by https://github.com/collectiveidea/awesome_nested_set/wiki/How-to-generate-nested-unordered-list-tags-with-one-DB-hit

Should I put these in a separate helper?

app/helpers/sheet/group/nav_left.rb
```diff
@@ -31,8 +31,31 @@ def render
...
+
       def groups
-        @groups ||= entry.groups_in_same_layer.without_deleted.order(:lft).to_a
+        @groups ||= sort_list_by_name entry.groups_in_same_layer.without_deleted.order(:lft)
```
rnestler, line 58: @mtnstar This implements the sorting for the groups, but not the separately listed sections. These seem to be sorted when changed, but behave a bit strange.

app/views/groups/_child_groups.html.haml
```diff
@@ -4,7 +4,7 @@
 -#  https://github.com/hitobito/hitobito.
 
 - @sub_groups.each do |label, groups|
-  = section_table(label, groups) do |item|
+  = section_table(label, groups.sort_by(&:sorting_name)) do |item|
```
rnestler, line 7: This may be a bit confusing for cases where the short name wouldn't be sorted the same as the full name, since `sorting_name` uses the short name:

image

So here `SAC Blüemlisalp` has the short name `SAC Z Blmli` and thus appears almost at the end.

app/helpers/sheet/group/nav_left.rb
```diff
@@ -31,8 +31,31 @@ def render
...
+      def recombine_groups(results, children_of, parent_id)
+        children_of[parent_id]&.each do |o|
+          results << o
+          recombine_groups(results, children_of, o.id)
+        end
+      end
+
       def groups
-        @groups ||= entry.groups_in_same_layer.without_deleted.order(:lft).to_a
+        @groups ||= sort_groups_by_name entry.groups_in_same_layer
```
rnestler, line 58: We could avoid sorting in cases where non of the groups have a translated `static_name`. But I'm not sure if the performance impact is big enough for us to care.

app/views/groups/_child_groups.html.haml
rnestler: Also I tried to convert the partial into a view component to avoid to have too much logic in the template. But the most trivial conversion lead to rendering problems which I couldn't solve trivially, since a lot of other partials are involved.

app/helpers/sheet/group/nav_left.rb
rnestler: Fixed with https://github.com/hitobito/hitobito_sac_cas/pull/621

app/helpers/sheet/group/nav_left.rb
rnestler: Resolved offline: Not now. If something similar is needed we can still extract it.

Largo (commented):
nice!

PR #2711, opened 2024-06-25
ADDRESSES: Bugfixes and Adaptations

kronn:
See #2226

### DIFF
--- a/app/domain/person/address.rb
+++ b/app/domain/person/address.rb
@@ -1,6 +1,6 @@
 # frozen_string_literal: true
 
-#  Copyright (c) 2012-2021, CVP Schweiz. This file is part of
+#  Copyright (c) 2012-2024, CVP Schweiz. This file is part of
 #  hitobito_cvp and licensed under the Affero General Public License version 3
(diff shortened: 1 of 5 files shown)


app/domain/person/address.rb
```diff
@@ -39,10 +39,12 @@ def combine_household_names(members)
 
   def address
     [
+      @person.address_care_of.to_s.strip.presence,
       @person.address.to_s.strip,
+      @person.postbox.to_s.strip.presence,
       [@person.zip_code, @person.town].compact.join(' ').squish,
       country
```
rnestler, line 46: Is `country` always present?

kronn: tl;dr yes

The method `country` (a few lines below this one in the same class) abstracts this nicely, I think. If it's the default-country (as set in `config/settings.yml`), it is nil and this line is compacted away, otherwise, the country is rendered.

If it is indeed an empty string, it will also be rendered needlessly in the PDF. Luckily, whitespace does not waste too much ink/toner. :-)

rnestler (approved):

PR #2723, opened 2024-07-03
Feature/606 Make wizard step accessor public

rnestler:
Refs: https://github.com/hitobito/hitobito_sac_cas/issues/606


rnestler:
@daniel-illi @amaierhofer I'll merge without review, since we discussed this and you agreed.

PR #2743, opened 2024-07-14
Reduce flaky specs by no altering the environment too much

kronn:
This has been tested by running the specs in the SAC_CAS-Wagon in a certain order:

rspec ./spec/decorators/role_decorator_spec.rb ./spec/resources/group/reads_spec.rb --seed 11597

Surely, this might happen anywhere where both generated URLs and decorators with draper are being tested.

### DIFF
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -115,7 +115,9 @@ RSpec.configure do |config|
 
   config.before(:each, :draper_with_helpers) do
     c = ApplicationController.new
-    c.request = ActionDispatch::TestRequest.new({})
+    request = ActionDispatch::TestRequest.new({})
+    request.host = "test.host" # ensure to not remove the test-host
+    c.request = request
     allow(c).to receive(:current_person) { people(:top_leader) }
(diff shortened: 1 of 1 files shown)


spec/spec_helper.rb
```diff
@@ -130,7 +130,9 @@
 
   config.before(:each, :draper_with_helpers) do
     c = ApplicationController.new
-    c.request = ActionDispatch::TestRequest.new({})
+    request = ActionDispatch::TestRequest.new({})
+    request.host = "test.host"
+    c.request = request
```
rnestler, line 135:
```suggestion
    c.request = ActionDispatch::TestRequest.new({host: "test.host"})
```

Doesn't work?

Also consider adding a comment explaining why this is needed.

kronn: this was my first try and that did not work. It seems that the passed ENV ist parsed differently than the exposed attribute API.

kronn: the most thorough comment is in the commit-message, a short version has been added to the most confusing line

PR #2748, opened 2024-07-23
MAIL: Localized Sendername (hitobito/hitobito_sac_cas#600)

Largo:


spec/mailers/application_mailer_spec.rb
```diff
@@ -0,0 +1,197 @@
...
+          de: {settings: {email: {sender: "de broken email"}}},
+          fr: {settings: {email: {sender: "fr broken email"}}}
+        ) do
+          example.call
+        end
+      end
+
+      it "has the sender from the settings if the locale changes" do
+        check_sender(->(message) { expect(message[:from].value).to eq(Settings.email.sender) }) do
+          InvoiceMailer.notification(invoice, sender)
```
Largo, line 176: I'm not sure how mail[:from].value here is not "de broken email". It's good, but I can't figure out how this happens. Mail:Message.from= does not do that.

TheWalkingLeek (approved):

PR #2753, opened 2024-07-30
Feature/2662 Small fixes in wizard base and steps_component

rnestler:
The smaller fixes extracted from https://github.com/hitobito/hitobito/pull/2750, since I'll need them as well.


rnestler:
Since #2665 got merged including these fixes, this can be closed.

PR #2759, opened 2024-08-04
Restrict access to query endpoint

carlobeltrame:
Refs https://help.puzzle.ch/#ticket/zoom/8081

### DIFF
--- a/app/abilities/person_ability.rb
+++ b/app/abilities/person_ability.rb
@@ -94,7 +94,7 @@ class PersonAbility < AbilityDsl::Base
   end
 
   def if_any_writing_permissions_or_any_leaded_events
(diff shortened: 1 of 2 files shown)


app/abilities/person_ability.rb
```diff
@@ -92,6 +93,10 @@ def if_any_writing_permissions
     contains_any?(writing_permissions, user_context.all_permissions)
   end
 
+  def if_any_writing_permissions_or_any_leaded_events
+    if_any_writing_permissions || user_context.events_with_permission(:event_full).present?
```
kronn, line 97: I am unsure if `.any?` would be better here. I tend to prefer ruby builtin methods over rails additions. My preference might be wrong here...

kronn (approved):

PR #2768, opened 2024-08-13
Feature/replace custom sequence

MarcEgliP:


spec/domain/sequence_spec.rb
```diff
@@ -0,0 +1,29 @@
...
+  end
+
+  context "::increment!" do
+    it "increments the current_value of the sequence" do
+      Sequence.increment!("test_sequence")
+      Sequence.increment!("test_sequence")
+      expect(Sequence.current_value("test_sequence")).to eq 3
+    end
+  end
+end
```
njaeggi, line 29: add a spec where start 10 and check if value is 11 after first increment, just in case

njaeggi (changes requested):
LGTM

Last thing to check is if it's possible to run the migrations, from the values in the specs it looks like household keys contained numbers, which would break the migration to sequence... Also UUID's look something like this: 3bbee048-ecc3-418b-bf5f-aa3d8a1b28cc

One small spec to add too :rocket:

PR #2783, opened 2024-08-23
Fix feature specs when running in Docker

rnestler:
We build the development assets and copy them to the test assets for the tests. The environment variable seemed to be set such that the test assets were built, but then overridden by the previous development assets.


openscript (commented):
It works for me too, but it feels wrong to change it there, because for testing it should run with the `test` env, doesn't it?

rnestler:
> It works for me too, but it feels wrong to change it there, because for testing it should run with the test env, doesn't it?

Well yes, but for some reason we need to compile the development assets and just copy them to the test assets. If we look at `config/webpacker.yml` it should be enough to just compile the test assets.

If I switch it to compile the `test` assets directly it fails:
```diff
 rm -rf public/packs-test/ && echo "remove public/packs-test"
-RAILS_ENV=development rake webpacker:compile
-cp -rf public/packs public/packs-test && echo "copied public/packs -> public/packs-test"
-
+RAILS_ENV=test rake webpacker:compile
```

If we look at 2b4807fa7eb289696f7966b15cdee46404514361 there was hope that this is just a temporary workaround :wink:

kronn (approved):
This is anyway the RAILS_ENV I run the script in. The copy assume that the generated pack is created in development or production-mode (i.e. not test).

I was told (back then) that this should work automatically in tests anyway and I am at fault.

Glad to see that my temporary fix stood the test of time :-)

rnestler:
> I was told (back then) that this should work automatically in tests anyway and I am at fault.

I'm wondering as well why the test assets are broken...

PR #2798, opened 2024-08-30
Avoid db access when app is loaded

codez:
Otherwise, tasks that must work without database would fail (e.g. rails db:create)


kronn (approved):

kronn:
I merged this manually into master, closing this PR

PR #2799, opened 2024-08-30
Stop Building sphinx-images

kronn:
There is no active development on the indices currently and the forseeable future, so there is no need to fill the registry with the ever-same image under a different version-number.


TheWalkingLeek (approved):

hunchr:
@TheWalkingLeek das kann gemerged werden, oder?

PR #2832, opened 2024-09-09
Make bin/setup usable

rnestler:
* Install JS dependencies
 * Run `db:create` and then `db:migrate wagon:migrate db:seed wagon:seed`, since `db:prepare` wouldn't run the wagon stuff


Largo:
LGTM

bin/setup
```diff
@@ -18,15 +18,16 @@ FileUtils.chdir APP_ROOT do
...
+  system!('bin/yarn')
 
   # puts "\n== Copying sample files =="
   # unless File.exist?('config/database.yml')
   #   FileUtils.cp 'config/database.yml.sample', 'config/database.yml'
   # end
 
   puts "\n== Preparing database =="
-  system! 'bin/rails db:prepare'
+  system! 'bin/rails db:create'
```
rnestler, line 29: An alternative would be to override the `db:prepare` task.

PR #2875, opened 2024-09-22
Fix event API leaders relation

### LINKED ISSUE #2848: API: 'include=leaders' is broken for `/events` API
### Aktuelles Verhalten

(when signed into pbs.puzzle.ch)

`https://pbs.puzzle.ch/api/events?include=dates,advisor&filter[group_id][eq]=391&filter[type]=Event::Course` returns HTTP 200 with advisor and dates fields included

`https://pbs.puzzle.ch/api/events?include=dates,advisor,leaders&filter[group_id][eq]=391&filter[type]=Event::Course` returns HTTP 500

### Gewünschtes Verhalten

According to https://github.com/hitobito/hitobito/pull/2550, an array of course leaders should be included in the response

### Schritte zu reproduzieren

sign in and execute API requests as listed above

bug was first identified on June 4th, 2024

### hitobito version

2.2.3

### Organisation

PBS

when reproducing the error in a local dev setup, this is the start of the error message:
```
rails-1  | === Graphiti Debug
rails-1  | Rendering:
rails-1  | Took: 71.03ms
rails-1  | ERROR: NoMethodError: undefined method `leaders' for #<Event::Course id: 145, type: "Event::Course", number: "123", motto: "", cost: "", maximum_part
```
adding `attr_accessor :leaders` to the course model at https://github.com/hitobito/hitobito/blob/ae0a8793266e487e39268447e5aef2e778f59fea/app/models/event/course.rb#L61 resolves the 500 error but leaders is returned as null (even when course leader roles are present).

After some debugging it looks to me like the resolve method of CourseResource at https://github.com/hitobito/hitobito/blob/7db65d43d0914dabec5e2b9f29ea7b1d1c16318c/app/resources/event/course_resource.rb#L28-L30 is never executed and therefore the leaders attr accessor is missing. When adding a resolve method to the EventResource model it is invoked. I tried to dig into how the polymorphic setup is supposed to work but couldn't figure it out after reading through https://www.graphiti.dev/tutorial/step_9.

### ToDos

_No response_

simfeld:
Fixes #2848

- makes sure that `:leaders` atribute accessor exists for courses
- adds a constant defining the leader roles to be overridden in wagons

### DIFF
--- a/app/models/event/course.rb
+++ b/app/models/event/course.rb
@@ -59,6 +59,8 @@
 # A course is a specialised Event that has by default applications,
 # preconditions and may give a qualification after attending it.
 class Event::Course < Event
+  LEADER_ROLES = [Event::Role::Leader].map(&:sti_name)
+
   # This statement is required because this class would not be loaded otherwise.
   require_dependency "event/course/role/participant"
 
--- a/app/resources/person/name_resource.rb
+++ b/app/resources/person/name_resource.rb
@@ -21,7 +21,7 @@ class Person::NameResource < ApplicationResource
         .joins(event_participations: :roles)
         .select("people.id, first_name, last_name, event_id AS leads_course_id")
         .where(event_participations: {event_id: course_ids},
-          event_roles: {type: Event::Role::Leader.sti_name})
+          event_roles: {type: Event::Course::LEADER_ROLES})
     end
   end
 
--- a/app/resources/event/course_resource.rb
+++ b/app/resources/event/course_resource.rb
@@ -23,10 +23,4 @@ class Event::CourseResource < EventResource
   def base_scope
     Event::Course.all.accessible_by(index_ability).includes(:groups, :translations).list
   end
-
-  def resolve(scope)
-    scope.to_a.tap do |events|
-      events.each { |event| event.singleton_class.attr_accessor :leaders }
-    end
-  end
 end
--- a/app/resources/event_resource.rb
+++ b/app/resources/event_resource.rb
@@ -68,4 +68,10 @@ class EventResource < ApplicationResource
   def index_ability
     JsonApi::EventAbility.new(current_ability)
   end
+
+  def resolve(scope)
+    scope.to_a.tap do |events|
+      events.select { |event| event.type == Event::Course.sti_name }.each { |event| event.singleton_class.attr_accessor :leaders }
+    end
+  end
 end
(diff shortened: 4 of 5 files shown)


app/models/event/course.rb
```diff
@@ -59,6 +59,8 @@
 # A course is a specialised Event that has by default applications,
 # preconditions and may give a qualification after attending it.
 class Event::Course < Event
+  LEADER_ROLES = [Event::Role::Leader].map(&:sti_name)
```
kronn, line 62: Since this is supposed to be overridden in Wagons, I would remove the `map` and do this when reading/using this list.

This would then also fit better with the `Event.role_types` which are also stored as constants.

app/resources/person/name_resource.rb
```diff
@@ -21,7 +21,7 @@ class Person::NameResource < ApplicationResource
         .joins(event_participations: :roles)
         .select("people.id, first_name, last_name, event_id AS leads_course_id")
         .where(event_participations: {event_id: course_ids},
-          event_roles: {type: Event::Role::Leader.sti_name})
+          event_roles: {type: Event::Course::LEADER_ROLES})
```
kronn, line 24:
```suggestion
          event_roles: {type: Event::Course::LEADER_ROLES.map(&:sti_name)})
```

app/models/event/course.rb
kronn, line 62:
```suggestion
  LEADER_ROLES = [Event::Role::Leader]
```

kronn (changes requested):

codez:
I just discovered this PR now. The behavior should be fixed with [!3019](https://github.com/hitobito/hitobito/pull/3019)

kronn:
I assume this has been fixed by #3019, therefore I close this PR. Judging from the time both PRs happened: Great minds think alike :-)

simfeld:
@kronn @codez unfortunately #3019 is only half of the solution. Yes there are no more HTTP 500 errors, but

- `course_leader_role` in the name_resource class still needs to be overridden in the pbs wagon
- as you can see in https://github.com/hitobito/hitobito_pbs/pull/343 the role model of PBS knows 2 roles you could consider to be course leaders ("Kursleiter\*in" and "Klassenlehrer\*in"), so it would be preferable not only have a singular course_leader_role` attribute

PR #2887, opened 2024-09-26
Run seeds in CI

rnestler:
In the SAC project we had the issue from time to time, that changes in the code would break the seeds and thus interrupt the workflow of other developers which needed to reseed locally.

To avoid that we want to run the seeds during CI so we'd notice of a PR would beeak the seeds.

Actual CI run is here: https://github.com/hitobito/hitobito_sac_cas/pull/1047

### DIFF
--- a/.github/workflows/wagon-tests.yml
+++ b/.github/workflows/wagon-tests.yml
@@ -106,6 +106,7 @@ jobs:
       - name: 'Prepare'
         uses: ./.hitobito_core_repo/.github/actions/wagon-ci-setup
         with:
+          assets: false
           wagon_repository: ${{ inputs.wagon_repository }}
           wagon_dependency_repository: ${{ inputs.wagon_dependency_repository }}
           core_ref: ${{ inputs.core_ref }}
--- a/.github/workflows/wagon-tests.yml
+++ b/.github/workflows/wagon-tests.yml
@@ -78,7 +78,8 @@ jobs:
       RAILS_DB_PASSWORD: hitobito
       RAILS_DB_NAME: hitobito_development
       RAILS_ENV: development
-
+    # only run seeds on master
+    if: ${{ github.ref_name == 'master' }}
     services:
       postgres:
         image: postgres:16
--- a/.github/workflows/wagon-tests.yml
+++ b/.github/workflows/wagon-tests.yml
@@ -66,6 +66,56 @@ jobs:
         run: |
           bundle exec rake app:rubocop
 
+  wagon_seeds:
+    name: seeds
+    runs-on: 'ubuntu-20.04'
+    env:
+      HEADLESS: true
+      RAILS_DB_ADAPTER: postgresql
+      RAILS_DB_HOST: 127.0.0.1
+      RAILS_DB_PORT: 5432
+      RAILS_DB_USERNAME: hitobito
+      RAILS_DB_PASSWORD: hitobito
+      RAILS_DB_NAME: hitobito_test
+      RAILS_TEST_DB_NAME: hitobito_test
+      RAILS_ENV: test
+
+    services:
+      postgres:
+        image: postgres:16
+        env:
+          POSTGRES_USER: hitobito
+          POSTGRES_PASSWORD: hitobito
+          POSTGRES_DB: hitobito_test
+        ports:
+          - '5432:5432'
+        options: >-
+          --health-cmd "pg_isready -U hitobito"
+          --health-interval 10s
+          --health-timeout 10s
+          --health-retries 10
+    steps:
+      - name: Check out the core, which contains the shared setup action
+        uses: actions/checkout@v4
+        with:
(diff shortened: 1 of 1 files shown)


.github/workflows/wagon-tests.yml
```diff
@@ -64,6 +64,55 @@ jobs:
...
+    env:
+      HEADLESS: true
+      RAILS_DB_ADAPTER: postgresql
+      RAILS_DB_HOST: 127.0.0.1
+      RAILS_DB_PORT: 5432
+      RAILS_DB_USERNAME: hitobito
+      RAILS_DB_PASSWORD: hitobito
+      RAILS_DB_NAME: hitobito_test
+      RAILS_TEST_DB_NAME: hitobito_test
+      RAILS_ENV: test
```
rnestler, line 79: Maybe we should run the seeds in the `development` environment?

.github/workflows/wagon-tests.yml
```diff
@@ -64,6 +64,55 @@ jobs:
...
+          path: .hitobito_core_repo
+          fetch-depth: 1
+
+      - name: 'Prepare'
+        uses: ./.hitobito_core_repo/.github/actions/wagon-ci-setup
+        with:
+          wagon_repository: ${{ inputs.wagon_repository }}
+          wagon_dependency_repository: ${{ inputs.wagon_dependency_repository }}
+          core_ref: ${{ inputs.core_ref }}
+          wagon_dependency_ref: ${{ inputs.wagon_dependency_ref }}
```
rnestler, line 110: Maybe we could skip this and just call `db:create`, `db:migrate`, `wagon:migrate`?

rnestler: We can just pass `assets: false` to speed it up.

.github/workflows/wagon-tests.yml
Largo: yes, I would say so.

kronn (commented):
Running the dev-seeds would lengthen the CI-Run significantly and most likely break a lot of specs.

Fixtures should be used for most cases.

For a generic solution to have things like custom contents in the test-setup, the architects should be consulted. I will request their review.

rnestler:
> Running the dev-seeds would lengthen the CI-Run significantly and most likely break a lot of specs.

I noticed that I didn't mention the intent of the change, which I now fixed. It's a separate step to verify that the seeds run sucessfully.

codez (approved):
I approve the idea of running the seed Rake task in CI to check that it works as expected (unrelated to specs as @rnestler mentioned above). However, running it on each PR would lengthen the actions too much. I suggest to only run it on the `master` branch (or on a periodic "nightly" run).

.github/workflows/wagon-tests.yml
```diff
@@ -64,6 +64,56 @@ jobs:
...
+    env:
+      HEADLESS: true
+      RAILS_DB_ADAPTER: postgresql
+      RAILS_DB_HOST: 127.0.0.1
+      RAILS_DB_PORT: 5432
+      RAILS_DB_USERNAME: hitobito
+      RAILS_DB_PASSWORD: hitobito
+      RAILS_DB_NAME: hitobito_test
+      RAILS_TEST_DB_NAME: hitobito_test
+      RAILS_ENV: test
```
kronn, line 79: Since we mostly have special seeds in den dev-env, we need to use that env to run them in this step.

```suggestion
      RAILS_ENV: development
```

Another option would be symlinking or copying the seeds from `db/seeds/development` to `db/seeds/test`.

kronn (changes requested):

kronn (approved):

.github/workflows/wagon-tests.yml
rnestler: Fixed.

.github/workflows/wagon-tests.yml
```diff
@@ -66,6 +66,56 @@ jobs:
...
+      HEADLESS: true
+      RAILS_DB_ADAPTER: postgresql
+      RAILS_DB_HOST: 127.0.0.1
+      RAILS_DB_PORT: 5432
+      RAILS_DB_USERNAME: hitobito
+      RAILS_DB_PASSWORD: hitobito
+      RAILS_DB_NAME: hitobito_development
+      RAILS_ENV: development
+    # only run seeds on master
+    if: ${{ github.ref_name == 'run-seeds-in-ci' }}
```
rnestler, line 82: I experimented with various incantation of this condition to check if I can get it to execute when the https://github.com/hitobito/hitobito_sac_cas/pull/1047 branch triggers the workflow (thus currently `run-seeds-in-ci` later we'd change it to `master`). But it didn't work :( it just always skips...

rnestler: Ahh
image

> For pull requests, the format is <pr_number>/merge.

PR #2894, opened 2024-10-01
Fix CrudController#update

hunchr:


Largo:
Originaler Commit:

[hitobito/app/controllers/people_controller.rb at 719b66176a6bfe384940073664d7cbebcc83ce82 · hitobito/hitobito](https://github.com/hitobito/hitobito/blob/719b66176a6bfe384940073664d7cbebcc83ce82/app/controllers/people_controller.rb)

PR #2902, opened 2024-10-06
Avoid overflow in signup view (hitobito/hitobito_sac_cas#1031)

Largo:
refs (hitobito/hitobito_sac_cas#1031)

See the gray overflow to the right:
image


hunchr (approved):
LGTM

PR #2922, opened 2024-10-16
Check for action destroy in find_entry

njaeggi:


app/controllers/people_controller.rb
```diff
@@ -119,10 +119,10 @@ def self.sort_mappings_with_indifferent_access
 
   alias_method :group, :parent
 
-  # every person may be displayed underneath the root group,
+  # every person may be displayed underneath the next layer group,
```
carlobeltrame, line 122: Comment is misleading. What you are actually doing here is more closely described by: "every person may be displayed underneath ANY layer group". Is that really the solution to this problem? Doesn't this open security issues or create new problems?

kronn (approved):

PR #2929, opened 2024-10-18
Rework database config

kronn:
This PR covers 3 aspects:

1. it drops support for sqlite3, which is not usable or needed for local development anymore.
2. it updates a config setting to the current version
3. it supports some more settings for the postgresql-connection. 

The newly supported settings are
- lock_timeout
- statement_timeout

Both help us to stop overly long-running queries and therefore stabilize the application. The actual settings are then done on the server/cluster with environment variables.


daniel-illi (approved):
LGTM :+1:

PR #2974, opened 2024-11-04
Feature/2936/quicksearch indexing

njaeggi:
This Implementation improves our search speed from around ~ `(21740.6ms)` to  ~ `(120.5ms)`, when searching around 200'000 people with multiple phone_numbers, emails and social accounts.

### DIFF
--- a/.github/actions/core-ci-setup/action.yml
+++ b/.github/actions/core-ci-setup/action.yml
@@ -78,5 +78,5 @@ runs:
     - name: 'Run db migrations'
       if: inputs.migrations == 'true'
       run: |
-        bundle exec rake db:migrate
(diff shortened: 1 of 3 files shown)


app/models/group.rb
```diff
@@ -56,7 +56,7 @@
 
 class Group < ActiveRecord::Base # rubocop:disable Metrics/ClassLength
   SEARCHABLE_ATTRS = [:name, :short_name, :email, :address, :zip_code, :town, :country,
-    {parent: [:name, :short_name], phone_numbers: [:number],
```
njaeggi, line 59: Searching groups based on the parent didn't work, with `PostgreSql` or `MySql`, if this is really something we need, we have to use a custom join query in our search method

PR #2980, opened 2024-11-05
Allow FormFieldCheckDependentController to check all dependent boxes

rnestler:
Configured with a stimulus value:
https://stimulus.hotwired.dev/reference/values

### DIFF
--- a/app/javascript/controllers/form_field_check_dependent_controller.js
+++ b/app/javascript/controllers/form_field_check_dependent_controller.js
@@ -7,12 +7,19 @@ import { Controller } from "@hotwired/stimulus";
 
 export default class extends Controller {
   static targets = ["main", "dependent"];
+  static values = {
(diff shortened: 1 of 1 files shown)


app/javascript/controllers/form_field_check_dependent_controller.js
rnestler, line 1: I couldn't find any existing spec for this in core. I guess it isn't used at all in core?

njaeggi: There is no use in core for this, the only use is in the SAC wagon, specs right here: https://github.com/hitobito/hitobito_sac_cas/blob/master/spec/features/memberships/sac_terminate_membership_wizard_spec.rb#L70-L109

rnestler: Then we can merge it like this.

njaeggi (approved):
LGTM

PR #3056, opened 2024-12-02
Switch to Ruby's own Enumerable#sum

kronn:
This resolves the deprecation warning about this as well.

### DIFF
--- a/app/models/invoice.rb
+++ b/app/models/invoice.rb
@@ -184,7 +184,7 @@ class Invoice < ActiveRecord::Base
 
   def calculated
     [:total, :cost, :vat].index_with do |field|
-      round(invoice_items.reject(&:frozen?).sum(&field))
+      round(invoice_items.reject(&:frozen?).map(&field).compact.sum(BigDecimal("0.00")))
     end
   end
 


app/models/invoice.rb
```diff
@@ -184,7 +184,7 @@ def last_payments_information
 
   def calculated
     [:total, :cost, :vat].index_with do |field|
-      round(invoice_items.reject(&:frozen?).sum(&field))
+      round(invoice_items.reject(&:frozen?).map(&field).compact.sum(BigDecimal("0.00")))
```
rnestler, line 187: This looks quite a bit more complex than before. Also I'm not sure it is needed, since `sum(&:acc)` works also in plain ruby without rails.

kronn: I agree. The map/compact addition is needed to not trip over nil-values. The much shorter version

```suggestion
      round(invoice_items.reject(&:frozen?).sum(0, &field))
```

also resolves the deprecation, but cannot handle nil-values. Both `0` and `BigDecimal("0.00")` behave the same. 

Therefore, chose the better type that we use for rounding anyway. The increased allocations are shortlived and therefore an easy target for the GC.

A lot of the perceived complexity was hidden in the Rails-version/Core-Ext of `Enumerable#sum`.

rnestler (approved):

PR #3063, opened 2024-12-04
Update rake tasks hitobito:roles, print role types

daniel-illi:


kronn:
Might I ask, why? The output is used for the README and was until now human-readable. Now it leaks internal details that are mostly only useful for devs. Would it be sensible to create a second rake-task for this?

PR #3068, opened 2024-12-05
The order of things is not checked here

kronn:
If it were, it would not fail from time to time.


rnestler (approved):
Assuming it's OK that it's not ordered this LGTM.

kronn:
> Assuming it's OK that it's not ordered this LGTM.

The expected order does not make sense to me, so I assume it is just what happens most of the time. If this is wrong, it has been for such a long time that it is now a feature that I wouldn't change whilly-nilly.

rnestler:
> The expected order does not make sense to me, so I assume it is just what happens most of the time. If this is wrong, it has been for such a long time that it is now a feature that I wouldn't change whilly-nilly.

I guess the order was fixed with MySQL and go non-deterministic with the change to Postgres.

PR #3072, opened 2024-12-09
Adjust some view helpers to support further options

njaeggi:


app/helpers/standard_form_builder.rb
```diff
@@ -287,7 +287,8 @@ def has_many_field(attr, html_options = {}) # rubocop:disable Naming/PredicateNa
     ].compact.join(" ")
     html_options[:class] += " is-invalid" if errors_on?(attr)
 
-    add_css_class(html_options, "multiselect")
+    add_css_class(html_options, "multiselect tom-select")
+    html_options[:data] = {chosen_no_results: I18n.t("global.chosen_no_results")}
```
njaeggi, line 291: For multi selects we have TomSelect, a tom select is initialized on every element with the `tom-select` class. For some reason the StandardFormBuilder did not add this class for `has_many_fields`

rnestler: Doesn't this change a lot of things in all hitobito instances if we add this to every affected form?

rnestler, line 291: `html_options[:data] = {chosen_no_results: I18n.t("global.chosen_no_results")}` doesn't seem to have any effect. Also I can't see the message `chosen_no_results: Keine Einträge gefunden mit` from the locale anywhere if it is set. Where should it show up?

njaeggi: Yes it does... kinda, as far as last time I upgraded bootstrap from 2 to 5 I didn't find a lot of multiselect fields in all wagons or core Features anyways.

Additionally this only changes the multiselect fields to look good, and does not really change any of the behaviour. All other options tom-select offers like search inside the multi select fields and so on, are not initialized by default in out implementation.

With TomSelect:
image

Without TomSelect:
image

njaeggi: It is used when the search is enabled for a multiselect, and no results are found, like in this Case:
image

But since we do not enable the search per default in this case anyways, you are correct, I can remove it :upside_down_face:

rnestler (approved):

PR #3074, opened 2024-12-10
Feature/1334/date picker format

njaeggi:
I tried adding a feature spec but the problem is that the datepicker isn't initialized in the spec I suppose, everywhere else we just use` fill_in ..., with: "12.12.2000"` which works fine, but the "validation/allowed keys" for the locales are checked in a keydown event, from the datepicker, `send_keys` also does not trigger the event, because it doesn't exist


app/javascript/javascripts/modules/datepicker.js.coffee
```diff
@@ -30,12 +30,14 @@ class app.Datepicker
     maxDate = if input.attributes.maxdate? then new Date(input.attributes.maxdate.value) else null
 
     options = $.extend({ onSelect: (d, i) -> self.track(this, d, i) },
+      $.datepicker.regional[$('html').attr('lang')],
```
rnestler, line 33: The language in `[$('html').attr('lang')]` is set to "fr", so I tried to hard-code it as `fr_CH` but it didn't work. So apparently `datepicker` doesn't support it.

rnestler (approved):
> I tried adding a feature spec but the problem is that the datepicker isn't initialized in the spec I suppose, everywhere else we just use fill_in ..., with: "12.12.2000" which works fine, but the "validation/allowed keys" for the locales are checked in a keydown event, from the datepicker, send_keys also does not trigger the event, because it doesn't exist

For sure there is an existing feature spec where datepicker is used where we could just switch the language to French, no?

But I tested it and it works.

app/javascript/javascripts/modules/datepicker.js.coffee
rnestler: Ah after importing `datepicker-fr-CH` it works with `fr-CH`:
```diff
diff --git a/app/javascript/packs/application.js b/app/javascript/packs/application.js
index 3bbcce375..e09d1031e 100644
--- a/app/javascript/packs/application.js
+++ b/app/javascript/packs/application.js
@@ -29,6 +29,7 @@ import 'moment'; // used by events/date_period_validator.js.coffee
 import 'jquery-ui/ui/widgets/datepicker';
 import 'jquery-ui/ui/i18n/datepicker-de';
 import 'jquery-ui/ui/i18n/datepicker-fr';
+import 'jquery-ui/ui/i18n/datepicker-fr-CH';
 import 'jquery-ui/ui/i18n/datepicker-it';
 import 'jquery-ui/ui/effects/effect-highlight';
 ```

File: app/javascript/packs/application.js [hunk 2]
```diff
@@ -29,7 +29,9 @@ import 'moment'; // used by events/date_period_validator.js.coffee
 import 'jquery-ui/ui/widgets/datepicker';
 import 'jquery-ui/ui/i18n/datepicker-de';
 import 'jquery-ui/ui/i18n/datepicker-fr';
+import 'jquery-ui/ui/i18n/datepicker-fr-CH';
 import 'jquery-ui/ui/i18n/datepicker-it';
+import 'jquery-ui/ui/i18n/datepicker-it-CH';
```
rnestler, line 34: I'm not sure if we could also remove the `fr` and `it` locales completely.

PR #3079, opened 2024-12-12
Bug/1392/password reset without locale

njaeggi:


app/controllers/devise/hitobito/passwords_controller.rb
```diff
@@ -19,7 +19,11 @@ def successfully_sent?(resource)
   def create
     previous_locale = I18n.locale
     resource = resource_class.find_by(email: resource_params["email"])
-    I18n.locale = resource&.language || previous_locale
+    I18n.locale = if Settings.application.languages.to_h.key?(resource&.language&.to_sym)
```
rnestler, line 22: ~~Shouldn't we look if the locale is in `I18n.available_locales`?~~ No, since `Settings.application.languages` is a subset of it.

rnestler, line 22:
```suggestion
    I18n.locale = if Settings.application.languages.keys.include?(resource&.language&.to_sym)
```

But doesn't matter much.

rnestler (approved):

PR #3085, opened 2024-12-16
Fix profile pictures on person view

rnestler:
This fixes the `upload_url` helper to extract the image dimensions as integers. With the migration to Rails 7.0 the Active Storage variant processor was changed to `:vips` which raised an "no implicit conversion
of String into Integer" exception when encountering the stringified integers. See also
https://guides.rubyonrails.org/upgrading_ruby_on_rails.html#active-storage-default-variant-processor-changed-to-vips


kronn (approved):
Any reasonable spec would need to a integration-spec which is IMHO pointless as this changes very seldomly. And it was caught in the integration environment.

PR #3105, opened 2024-12-23
Confirm adding participant on application market

TheWalkingLeek:
Refs: https://github.com/hitobito/hitobito_sac_cas/issues/1389

### DIFF
--- a/app/views/event/application_market/_application.html.haml
+++ b/app/views/event/application_market/_application.html.haml
@@ -5,11 +5,7 @@
 
 %tr{id: dom_id(p)}
   %td
-    = link_to(icon('arrow-left'),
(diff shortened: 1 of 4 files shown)


app/views/event/application_market/_application.html.haml
```diff
@@ -8,7 +8,8 @@
     = link_to(icon('arrow-left'),
               participant_group_event_application_market_path(@group, @event, p),
               remote: true,
-              method: :put)
+              method: :put,
+              data: { confirm: t('.add_participant_confirm') })
```
rnestler, line 12: Isn't there any feature spec that now would need to click the confirm button?

PR #3116, opened 2025-01-06
Small documentation updates

brknkfr:
- Fix url to [Deployment Instructions](https://github.com/hitobito/hitobito/blob/master/doc/operator/01_deployment.md)
- Add note about minimal requirement for postgresql database (version 16 because of unicode collation, see https://github.com/hitobito/hitobito/issues/3114)


kronn:
Thanks for this.

PR #3133, opened 2025-01-10
Use new syntax to initialize IMAP

codez:


amaierhofer:
Obsolete, change mittlerweile schon gemerged

PR #3147, opened 2025-01-16
Add exlude_attr method to table displays column (hitobito/hitobito_sac_cas#1512)

njaeggi:
This Pull Request introduces #exclude_attr? which was an SAC implementation, to not display certain table display attributes in some groups.

This came with the issue that when a person selects an attribute, that does not exist in another group, the query is still trying to execute with the "excluded attr". That throws an error which is also fixed in this Pull Request (all columns that are `exclued_attr? = true`, will not be displayed/exported in any way, also when selected via a previous group/condition where it was possible)


daniel-illi (approved):
LGTM :+1:

PR #3162, opened 2025-01-23
Some organizations have multiple leader-roles

### LINKED ISSUE #2848: API: 'include=leaders' is broken for `/events` API

kronn:
fixes #2848, #2875

See also #3019

### DIFF
--- a/app/resources/person/name_resource.rb
+++ b/app/resources/person/name_resource.rb
@@ -9,8 +9,8 @@ class Person::NameResource < ApplicationResource
   self.model = ::Person
   self.type = "person-name"
 
-  class_attribute :course_leader_role
-  self.course_leader_role = Event::Role::Leader
+  class_attribute :course_leader_roles
+  self.course_leader_roles = [Event::Role::Leader]
 
   with_options writable: false, filterable: false, sortable: false do
     attribute :first_name, :string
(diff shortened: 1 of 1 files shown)


simfeld:
Thank you!

PR #3170, opened 2025-01-31
Remove unneeded notification exception

kronn:
The MailProcessedBeforeError was a signal that needed action. In the current mail-stack the duplicate mail is being moved to "failed" and therefore not processed again anymore. Since there is nothing to be done by an operator or developer, this only is a signal. The fact that this is not the responsibility of the people running, but the people owning the the instance reduces this signal to mere noise. Unneeded noise in a world that is too loud anyway.

Therefore, and not at all with a heavy heart, I joyfully propose to remove this piece of legacy that has been dutyfully ported into the present. I could go on discussing the merits or faults of the removed code, but few will remember and most likely no one will miss it.

https://sentry.puzzle.ch/pitc/hitobito-backend/issues/74204
https://sentry.puzzle.ch/pitc/hitobito-backend/issues/74285


njaeggi (changes requested):
Die Entfernung der Exception ist sinnvoll. Der CI schlägt aktuell noch fehl, bitte noch beheben. 😊

PR #3174, opened 2025-02-04
Add validation options for scale and percision of decimal values in mounted attrs (hitobito/hitobito_sac_cas#1590)

njaeggi:


app/models/concerns/mounted_attr.rb
```diff
@@ -122,6 +122,15 @@ def define_mounted_attr_validations(config)
           end
         end
 
+        if config.attr_type == :decimal
```
TheWalkingLeek, line 125: I think that warrants some specs :woozy_face:

TheWalkingLeek (changes requested):

PR #3186, opened 2025-02-11
Structure parser Part II

kronn:
The current improvement round offers the following new features:

- nesting of layers (Root < Sublayer)
- validation of used permissions (no more typos)
- allow forcing the wanted class-name for roles (" -- (Group::SubGroup::RoleName)"-notation)
- direct creation of files

Also, there is now a DRY_RUN for the rake-task.
For consistency, the output of `hitobito:roles` has been extended. The role-classes are there optionally available via `hitobito:roles[true]`.

Some next steps are also shown, mostly so that I do not forget. Working with the new structure does not end with this, it only starts with this task.

### DIFF
--- a/app/domain/structure_parser.rb
+++ b/app/domain/structure_parser.rb
@@ -94,10 +94,13 @@ class StructureParser
     end
 
     def class_name
-      @class_name ||= @name.encode(
-        "ASCII", "UTF-8",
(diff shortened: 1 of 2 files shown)


daniel-illi (approved):
LGTM :+1:

PR #3219, opened 2025-02-26
Use Arel.sql to prevent UnknownAttributeReference

njaeggi:
fixes: https://sentry.puzzle.ch/pitc/hitobito-backend/issues/75046

### DIFF
--- a/config/brakeman.ignore
+++ b/config/brakeman.ignore
@@ -22,6 +22,29 @@
     {
       "warning_type": "SQL Injection",
       "warning_code": 0,
+      "fingerprint": "39e07376e15a6bffb01f4a571fd6a089f4872cf483902bfc66190c97919a357c",
+      "check_name": "SQL",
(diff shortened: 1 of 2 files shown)


config/brakeman.ignore
```diff
@@ -19,6 +19,29 @@
...
+        "type": "method",
+        "class": "Event::Filter",
+        "method": "list_entries"
+      },
+      "user_input": "sort_expression.keys[0].gsub(/\\b\\w+\\./, \"\")",
+      "confidence": "Medium",
+      "cwe_id": [
+        89
+      ],
+      "note": "No SQL Injection possible here, because no user data is handled"
```
carlobeltrame, line 43: Are you absolutely sure? E.g. in https://github.com/hitobito/hitobito/blob/master/app/controllers/events_controller.rb#L262 a user-supplied query parameter is passed through the job to the filter, no?

PR #3221, opened 2025-02-26
Update local developer doc with regard to test_schema_maintenance

amaierhofer:


bin/active_wagon.rb
```diff
@@ -89,7 +89,7 @@ def environment(core: true)
       export RAILS_DB_USERNAME=hitobito
       export RAILS_DB_PASSWORD=hitobito
       export RAILS_DB_NAME=hit_#{wagon}_dev
-      export RAILS_TEST_DB_NAME=hit_#{wagon}_test
+      export RAILS_TEST_DB_NAME=hit_#{core ? "core" : wagon}_test
```
kronn, line 92: Is it possible to also cover "dependency" wagons here? So far, it is only "youth" and "tenants", but logically, the test-DB should always be the one from the current directory to support local testing.

doc/developer/local_setup.md
```diff
@@ -28,15 +28,27 @@ Install all ruby and node dependencies (in the core folder):
 
 In the core directory:
 
-    ./bin/wagon activate generic
+    ./bin/active_wagon.rb generic
```
kronn, line 31: we could rename `bin/active_wagon.rb` to `bin/activate_wagon`, right? with the extension it feels like a library. For any executable, the extension should not matter as the `#!/path/to/runtime` is present.

kronn (changes requested):

doc/developer/local_setup.md
amaierhofer: Sure :+1:

PR #3233, opened 2025-03-05
When deleting addresses, do not validate

kronn:
We know that this is desctructive, which is why there is a report to STDOUT, via mail and an a PaperTrail associated with the affected person-entry in the database.

See #3144


amaierhofer (approved):
:+1:

PR #3242, opened 2025-03-13
add new develop

hoangkhanh1003:


ThomasEllenberger:
Hi

We are happy to see that you want to use, and further improve Hitobito.

The repository you have selected is used for the core of Hitobito and
therefore used by many other organizations. Organization-specific things
like the logo or custom models should go into a wagon (which is mostly
like a rails-engine). I suggest you fork the [generic wagon](https://github.com/hitobito/hitobito_generic)
as a starting point. You can (and should) rename it to match your
organzation to make it your own. If you want, you can take inspiration
from other hitobito_*-repositories to get a better feeling for this. In
this wagon, you can do whatever you need without interfering with the
core functionality.

If you find a bug in Hitobito itself, PRs are welcome. If you have a new
feature that you think everyone needs, PRs are also welcome. In general,
it is better to first develop in your wagon to fully support your
organization.

Thank you for improving and using Hitobito for your organization. Please
remember to keep Hitobito as AGPL3, as this will hopefully help the next
person (also, it is required 😉).

PR #3260, opened 2025-03-22
Update _footer.html.haml

richardjubla:
Simple and quick solution so that the logo and the text OR the text without the logo is better displayed. solves the problem that the text is respnsively pushed into the logo.

image

respnsive


carlobeltrame (approved):
It's better than before. I tested it with all wagons that have a footer.logo_image set (Jubla, Cevi, JEMK). For the others, this code is unused anyways.

PR #3267, opened 2025-03-24
Fix bugs related to filtering dates

njaeggi:

PR #3269, opened 2025-03-24
Fix blank constraint handling for person filter

amaierhofer:


amaierhofer:
obsolete durch https://github.com/hitobito/hitobito/pull/3267

PR #3362, opened 2025-05-07
 Use to_s on additional_emails too in contactable decorator 

njaeggi:
https://sentry.puzzle.ch/pitc/hitobito-backend/issues/76084


amaierhofer:
replaced by #3363

PR #3377, opened 2025-05-08
Bounce Handling

kronn:


nerrehmit:
From my very limited ruby know-how and with looking mainly at the changelog  I have a couple of questions.
Feel free to disregard any comments that you deem inapplicable/ out-of-scope.

- Do you preserve the bounce message that was received (for example a full mailbox or any other temporary reason?)
- Do you somehow differentiate between hard and soft bounces? I have a fear that 5 (hard) bounces might be a bit much for lists that get mail fairly infrequently.
- Do you generate statistics/ metrics about bounce percentage of certain lists (or recipient domains)? This could be used for an internal alarming if issues start to creep up.
- Do you reset the bounce counter if an email can be sent without a problem?
- What do you plan to do to reactivate an email address that was blocked (after 5 bounces). Is this an admin action or do you plan to do a verification before an address can be reactivated.

spec/models/bounce_spec.rb
```diff
@@ -0,0 +1,201 @@
...
+    end
+
+    context "of a blocked email" do
+      it "is accepted" do
+        expect do
+          blocked.block!
+        end.to_not change { blocked.reload.blocked? }
+      end
+
+      it "returns the timestamp of the original blocking" do
```
njaeggi, line 182: Add a similar spec to check that `Bounce#record` also does not update blocked_at as soon as it has been blocked

app/models/bounce.rb
```diff
@@ -0,0 +1,86 @@
...
+    # between the param and the stored array.
+    where("mailing_list_ids && '{:mailing_list_id}'", mailing_list_id: id)
+  }
+
+  before_validation :evaluate_blocking
+
+  class << self
+    def record(email, mailing_list_id: nil)
+      bounce = find_or_create_by(email: email)
+      if mailing_list_id.present?
```
njaeggi, line 29: This duplicates the mailing_list_ids in the array... Idk if that is the desired behavior, but it seems odd:
`mailing_list_ids: [1, 1, 1, 1, 1]`

config/routes.rb
```diff
@@ -313,6 +313,7 @@
 
         resources :mailchimp_synchronizations, only: [:create]
         resources :recipient_counts, controller: 'mailing_lists/recipient_counts', only: [:index]
+        resources :bounces
```
njaeggi, line 316:
```suggestion
        resources :bounces, only: [:index]
```

app/controllers/bounces_controller.rb
```diff
@@ -0,0 +1,29 @@
...
+  def group
+    @group ||= Group.find(params[:group_id])
+  end
+
+  def mailing_list
+    @mailing_list ||= MailingList.find(params[:mailing_list_id])
+  end
+
+  def model_scope
+    model_class.of_mailing_list(@mailing_list.id)
```
njaeggi, line 27: Do we really wanna show all bounces on the mailing_list view, this currently shows all bounces, also the ones that aren't blocked...

njaeggi (changes requested):
Paar kleiner Dinge sind noch nicht ganz korrekt :smile:

app/models/wizards/register_new_user_wizard.rb
```diff
@@ -68,6 +68,7 @@ def enqueue_duplicate_locator_job
 
   def enqueue_notification_email
     return if group.self_registration_notification_email.blank?
+    return if Bounce.blocked?(group.self_registration_notification_email)
```
njaeggi, line 71: This can be removed since the mailer uses `custom_content_mail`

njaeggi (changes requested):

app/controllers/bounces_controller.rb
kronn: True. The idea is: you can see when a mail last did not arrive. for this, you need to know IF there is a bounce at all.

PR #3386, opened 2025-05-15
Use event participation as model_class for participation exports (#3353)

njaeggi:
We don't want person table attrs inside our participation query :exploding_head:


kronn (changes requested):
While I agree, I'd like to see an actual spec for the exported data. So far, we only verify that 3 lines are in the CSV. This change might change the export significantly, so I want to have a spec that shows it. The specs could be a separate branch that we can base the current master to verify the correctness (same-output-ness?) of the model-class change.

PR #3392, opened 2025-05-16
Avoid empty translations by reusing the source

kronn:
The mode 'onlytranslated' omits reviewed translations. Therefore, we use the mode 'sourceastranslation' to avoid empty translations.


amaierhofer (commented):
:+1: glaub das ist der richtige mode, ich merge das mal

PR #3397, opened 2025-05-16
Allow for unicode in encrypted passwords

kronn:
Sentry: https://sentry.puzzle.ch/pitc/hitobito-backend/issues/76101

The decrypted value have an unknown encoding. As we only accept utf-8 for input-values, the resulting value (that we stored ourselves) needs be set to utf-8.


njaeggi (approved):
LGTM

PR #3406, opened 2025-05-20
Register tooltip on turbo render (#3043)

Vakmeth:
Fixes: #3043

### DIFF
--- a/app/javascript/javascripts/modules/tooltips.js
+++ b/app/javascript/javascripts/modules/tooltips.js
@@ -3,12 +3,12 @@
 //  or later. See the COPYING file at the top-level directory or at
 //  https://github.com/hitobito/hitobito.
 
-import * as bootstrap from 'bootstrap'
-
 function registerTooltips() {
-  const tooltipTriggerList = [].slice.call(document.querySelectorAll('[data-bs-toggle="tooltip"]'))
+  const tooltipTriggerList = Array.from(document.querySelectorAll('[data-bs-toggle="tooltip"]'))
   tooltipTriggerList.forEach(el => {
-    new bootstrap.Tooltip(el)
+    new Tooltip(el, {
+      placement: 'right',
+    })
   })
 }
 
--- a/app/javascript/javascripts/modules/tooltips.js
+++ b/app/javascript/javascripts/modules/tooltips.js
@@ -1,19 +1,19 @@
-//  Copyright (c) 2025, Hitobito. This file is part of
+//  Copyright (c) 2025, Hitobito This file is part of
 //  hitobito and licensed under the Affero General Public License version 3
 //  or later. See the COPYING file at the top-level directory or at
 //  https://github.com/hitobito/hitobito.
 
-// enable tooltip on page load
-$(document).on('turbo:load', function() {
-  $('[data-bs-toggle="tooltip"]').tooltip({ placement: 'right'});
-});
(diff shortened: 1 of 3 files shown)


app/javascript/javascripts/modules/tooltips.js
```diff
@@ -5,4 +5,8 @@
 
 document.addEventListener('DOMContentLoaded', function() {
   $('[data-bs-toggle="tooltip"]').tooltip({ placement: 'right' });
-});
\ No newline at end of file
+});
+
+$(document).on('turbo:load', function() {
```
njaeggi, line 10: Can we add all events like we do with tom-select?  So the next person trying to add a tooltip to a popover doesn't have todo this all over again :smile: 
https://github.com/hitobito/hitobito/blob/master/app/javascript/javascripts/modules/tom_select.js


https://github.com/hitobito/hitobito/blob/master/app/javascript/controllers/tom_select_controller.js
Maybe you could also write a stimulus controller for this just like tom select. You don't have to switch every tooltip in the whole app to the controller: "tooltip" option, but it may be even cleaner when we implement it to use in the future.

njaeggi (changes requested):

app/javascript/javascripts/modules/tooltips.js
njaeggi: Nice! The change from "DOMContentLoaded" to "turbo:load" also doesn't break the `html: true` when loading a page via reload?

app/javascript/packs/application.js
```diff
@@ -33,6 +33,7 @@ import 'jquery-ui/ui/i18n/datepicker-it-CH';
 import 'jquery-ui/ui/effects/effect-highlight';
 
 // Bootstrap
+import 'bootstrap/dist/js/bootstrap.bundle'
```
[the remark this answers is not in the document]
njaeggi: Also importing the whole bundle doesn't make sense when we import single stuff like:
import 'bootstrap/js/src/alert'
import 'bootstrap/js/src/button'

app/javascript/javascripts/modules/tooltips.js
```diff
@@ -1,19 +1,19 @@
...
+//  Copyright (c) 2025, Hitobito This file is part of
 //  hitobito and licensed under the Affero General Public License version 3
 //  or later. See the COPYING file at the top-level directory or at
 //  https://github.com/hitobito/hitobito.
 
-// enable tooltip on page load
-$(document).on('turbo:load', function() {
-  $('[data-bs-toggle="tooltip"]').tooltip({ placement: 'right'});
-});
+import * as bootstrap from 'bootstrap'
```
njaeggi, line 6: Why do we need this here, when you add it in application.js as well?

njaeggi (changes requested):
Some questions and cleanup suggestions, also you removed the default placement: right

app/javascript/packs/application.js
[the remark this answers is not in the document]
Vakmeth: Yeah you're right, makes no sense to import the whole bundle. I changed the import statement to only use the Tooltip of bootstrap.

njaeggi (approved):
LGTM, still don't know why you added the import, and why it worked before without it, please explain why it is needed, if it really is, you can merge this (squash)

PR #3407, opened 2025-05-20
Validate merged roles before merging (#3396)

njaeggi:


app/controllers/person_duplicates/merge_controller.rb
```diff
@@ -9,10 +9,14 @@ module PersonDuplicates
   class MergeController < ApplicationController
     before_action :authorize_action
 
+    helper_method :entry
+
     def new
     end
 
     def create
+      return rerender_form(:unprocessable_entity) unless entry.valid?(context: :merge)
```
kronn, line 18: 1. Nice, I have never used this argument in my life (loki-never-saw-this-in-his-life.jpg). Finally a good use-case
2. I suggest to not pass a hash, but only a value...

```suggestion
      return rerender_form(:unprocessable_entity) unless entry.valid?(:merge)
```

I humbly think it reads well enough.

app/models/person_duplicate.rb
```diff
@@ -29,6 +29,8 @@ class PersonDuplicate < ActiveRecord::Base
 
   validates :person_1_id, uniqueness: {scope: [:person_2_id]}
 
+  validate :valid_role_operations, if: -> { validation_context == {context: :merge} }
```
kronn, line 32: following the shortening above, we need to shorten things here as well...

```suggestion
  validate :valid_role_operations, if: -> { validation_context == :merge }
```

spec/models/person_duplicate_spec.rb
```diff
@@ -24,22 +24,27 @@
...
 
-        duplicate.update!(person_1: higher_id_person, person_2: lower_id_person)
+    it "is invalid in merge context when there is any possible invalid role when merging" do
+      bottom_member.roles.first.update_columns(start_on: 10.days.ago, end_on: 20.days.ago)
+      expect(duplicate).to be_valid
 
-        expect(duplicate.person_1).to eq(lower_id_person)
-        expect(duplicate.person_2).to eq(higher_id_person)
-      end
+      expect(duplicate.valid?(context: :merge)).to be_falsey
```
kronn, line 46: Lastly, our expectations should also be shortened:

```suggestion
      expect(duplicate.valid?(:merge)).to be_falsey
```

kronn (approved):
I do have one change-request, but I will apply it myself. Therefore my general approval stands.

PR #3409, opened 2025-05-21
Order invoices when exporting as pdf (hitobito/hitobito_sww#237)

njaeggi:


app/controllers/invoices_controller.rb
```diff
@@ -128,7 +128,7 @@ def render_invoices_pdf(invoices)
       with_async_download_cookie(format, filename(format, invoices)) do |filename|
         Export::InvoicesJob.new(format,
           current_person.id,
-          invoices.pluck(:id),
```
njaeggi, line 131: `pluck` replaces the whole SELECT segment, which removes `DISTINCT ON` from our query, leading to the job being called with some id's twice

PR #3410, opened 2025-05-21
bug/2079-participation-application

Vakmeth:
Fixes: #2079

### DIFF
--- a/app/helpers/dropdown/people_login.rb
+++ b/app/helpers/dropdown/people_login.rb
@@ -36,6 +36,7 @@ module Dropdown
     def send_login
       if template.can?(:send_password_instructions, @user) &&
           @user.email && (@user.roles.any? || @user.root?)
+
         add_item(translate(".send_login"),
           template.send_password_instructions_group_person_path(template.parent, @user),
           method: :post,
--- a/app/javascript/javascripts/modules/tooltips.js
+++ b/app/javascript/javascripts/modules/tooltips.js
@@ -1,8 +1,19 @@
-//  Copyright (c) 2012-2017, Jungwacht Blauring Schweiz. This file is part of
+//  Copyright (c) 2025, Hitobito This file is part of
 //  hitobito and licensed under the Affero General Public License version 3
 //  or later. See the COPYING file at the top-level directory or at
 //  https://github.com/hitobito/hitobito.
 
-document.addEventListener('DOMContentLoaded', function() {
-  $('[data-bs-toggle="tooltip"]').tooltip({ placement: 'right' });
-});
+function registerTooltips() {
+  const tooltipTriggerList = Array.from(document.querySelectorAll('[data-bs-toggle="tooltip"]'))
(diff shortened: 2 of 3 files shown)


app/views/event/application_market/_participation.html.haml
```diff
@@ -4,15 +4,16 @@
 -#  https://github.com/hitobito/hitobito.
 
 %td
-  %strong= link_to(p.to_s(:list), group_event_participation_path(@group, @event, p.id))
+  -# %strong= link_to(p.to_s(:list), group_event_participation_path(@group, @event, p.id))
```
njaeggi, line 7: This probably wasn't on purpose

spec/views/event/application_market/_participation.html.haml_spec.rb
```diff
@@ -0,0 +1,15 @@
+#  Copyright (c) 2025, Hitobito AG. This file is part of
+#  hitobito and licensed under the Affero General Public License version 3
+#  or later. See the COPYING file at the top-level directory or at
+#  https://github.com/hitobito/hitobito.
+
+require "spec_helper"
+
+describe "event/application_market/_participation.html.haml" do
+  let(:event) { EventDecorator.decorate(Fabricate(:course, groups: [groups(:top_layer)])) }
```
njaeggi, line 9:
```suggestion
  let(:event) { Fabricate(:course, groups: [groups(:top_layer)]).decorate }
```

spec/views/event/application_market/_participation.html.haml_spec.rb
```diff
@@ -0,0 +1,15 @@
...
+  let(:participation) { Fabricate(:event_participation, event: event) }
+
+  it "participation without application isn't causing an internal server error" do
```
njaeggi, line 12: I would call the spec something like "displays participation without application data", instead of saying we don't want an error here

spec/views/event/application_market/_participation.html.haml_spec.rb
```diff
@@ -0,0 +1,15 @@
...
+    expect { render locals: {p: participation.decorate} }.not_to raise_error
```
njaeggi, line 13: Also for this, I would rather render the view and check the content (when content exists how you want it, there is no error). Add a check for the expected text like person_location_information or whatever else is rendered in the view

njaeggi (changes requested):
Nice view specs, some smaller stuff you could improve about the spec

spec/views/event/application_market/_participation.html.haml_spec.rb
Vakmeth: But this isn't the bug is it? The Bug is that we get a 500 Internal Server error so the best spec to check this is to verify if the rendering of the page causes any error. The check if the right content is displayed has nothing to do with the Bug I fixed.

njaeggi (approved):

PR #3413, opened 2025-05-23
bug/2274-delete-tags

Vakmeth:
Bug ist nicht mehr relevant durch die Änderung: https://github.com/hitobito/hitobito/blob/bdbd4b7a73420b87df38266ebe287e89a65d6c50/config/initializers/acts_as_taggable_on.rb#L7

Siehe: https://github.com/mbleigh/acts-as-taggable-on/blob/master/lib/acts-as-taggable-on/tagging.rb

Durch das Setzen von `ActsAsTaggableOn.remove_unused_tags = false` wird das Tag nach dem Entfernen des letzten Taggings nicht mehr gelöscht.

Zur Überprüfung dieses Verhaltens wurde ein Test hinzugefügt.

Fixes: #2274

### DIFF
--- a/spec/controllers/person/tags_controller_spec.rb
+++ b/spec/controllers/person/tags_controller_spec.rb
@@ -166,6 +166,24 @@ describe Person::TagsController do
       expect(bottom_member.tags.count).to eq(0)
       is_expected.to redirect_to group_person_path(bottom_member.groups.first, bottom_member)
     end
+
+    let!(:test_tag) { ActsAsTaggableOn::Tag.create!(name: "Test") }
+    let!(:test_tagging) { ActsAsTaggableOn::Tagging.create!(tag_id: test_tag.id, taggable_type: "Person", taggable_id: bottom_member.id, context: "tags") }
+    let!(:subscription_tag) { SubscriptionTag.create!(excluded: false, subscription_id: Subscription.first.id, tag_id: test_tag.id) }
+
+    it "does delete tag if no other person uses it" do
+      expect do
+        delete :destroy, params: {
+          group_id: bottom_member.groups.first.id,
+          person_id: bottom_member.id,
+          name: test_tag.name
+        }
+      end
+        .not_to raise_error
+
+      expect(bottom_member.tags.count).to eq(0)
+      is_expected.to redirect_to group_person_path(bottom_member.groups.first, bottom_member)
+    end
   end
 
   private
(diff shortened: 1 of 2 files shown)


spec/controllers/person/tags_controller_spec.rb
```diff
@@ -166,6 +171,20 @@
       expect(bottom_member.tags.count).to eq(0)
       is_expected.to redirect_to group_person_path(bottom_member.groups.first, bottom_member)
     end
+
+    it "does delete tag if no other person uses it" do
```
njaeggi, line 175: Why is the spec called "does delete tag ..." and your explanation is that id doesn't delete it with our option? Also this spec just checks if the taggable on the person (`expect(top_leader.tags.count).to eq(0)`) was destroyed, not if the tag still exists or not. The association may not exist anymore, but you never check if a tag was actually destroyed/not destroyed according to your explanation in this PR.

spec/controllers/person/tags_controller_spec.rb
```diff
@@ -135,6 +136,10 @@
   end
 
   describe "DELETE #destroy" do
+    let!(:subscription_tag) { SubscriptionTag.create!(excluded: false, subscription_id: Subscription.first.id, tag_id: test_tag.id) }
+    let!(:test_tagging) { ActsAsTaggableOn::Tagging.create!(tag_id: test_tag.id, taggable_type: "Person", taggable_id: top_leader.id, context: "tags") }
+    let!(:test_tag) { ActsAsTaggableOn::Tag.create!(name: "Test") }
```
njaeggi, line 141: I know that for "tag" we have fabricators, if we have fabricators for the other records too, please use fabricators over .create!. This removes a lot of code around the stuff happening before the actual spec.

Vakmeth: We could, for reasons of unity I will keep the creates here, since the other tags and taggables are created in the same matter in this controller test.

spec/controllers/person/tags_controller_spec.rb
```diff
@@ -166,6 +170,17 @@
       expect(bottom_member.tags.count).to eq(0)
       is_expected.to redirect_to group_person_path(bottom_member.groups.first, bottom_member)
     end
+
+    it "does not raise error when last taggable association of tag was deleted" do
```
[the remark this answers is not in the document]
Vakmeth: Because the test should not test if the tagging was deleted correctly, this is already tested in another case. This test should only test that the deletion of a tagging does not cause an error like it was the case in the error @daniel-illi expected.

njaeggi (approved):

PR #3415, opened 2025-05-23
bug/3090-event-registration-display

Vakmeth:
Fixes: #3090

### DIFF
--- a/app/models/event/participation_contact_data.rb
+++ b/app/models/event/participation_contact_data.rb
@@ -7,14 +7,17 @@ class Event::ParticipationContactData
   attr_reader :person
 
   class_attribute :mandatory_contact_attrs,
-    :contact_attrs,
-    :contact_associations
+                  :address_attrs,
+                  :contact_attrs,
+                  :contact_associations
 
   self.mandatory_contact_attrs = [:email, :first_name, :last_name]
 
+
+  self.address_attrs = [:street, :housenumber, :postbox, :zip_code, :town]
+
   self.contact_attrs = [:first_name, :last_name, :nickname, :company_name, :email,
-    :address_care_of, :street, :housenumber, :postbox, :zip_code, :town,
-    :country, :gender, :birthday, :phone_numbers, :language]
+    :address_care_of, :country, :gender, :birthday, :phone_numbers, :language] + address_attrs
 
   self.contact_associations = [:additional_emails, :social_accounts]
 
@@ -67,6 +70,10 @@ class Event::ParticipationContactData
     return person.send(method) if /^.*household_key/.match?(method)
 
     super
+  end
+
+  def show_address
+    (attribute_keys & address_attrs).any?
   end
 
   def show_attr?(a)
--- a/app/views/event/participation_contact_datas/_address_fields.html.haml
+++ b/app/views/event/participation_contact_datas/_address_fields.html.haml
@@ -4,14 +4,17 @@
 -#  https://github.com/hitobito/hitobito_pbs.
 
 = f.labeled_input_field(:address_care_of) if entry.show_attr?(:address_care_of)
-= f.labeled(:address, class: 'd-flex') do
-  %div.col-9
-    = f.input_field(:street, data: { provide: 'entity', updater: 'AddressTypeahead.update', url: addresses_query_url }) if entry.show_attr?(:street)
-  %div.col-2.offset-1
-    = f.input_field(:housenumber) if entry.show_attr?(:housenumber)
-= f.labeled_input_field(:postbox)
-= f.labeled_input_field(:zip_code) if entry.show_attr?(:zip_code)
-= f.labeled_input_field(:town) if entry.show_attr?(:town)
+
+- if entry.show_address
+  = f.labeled(:address, class: 'd-flex') do
+    %div.col-9
+      = f.input_field(:street, data: { provide: 'entity', updater: 'AddressTypeahead.update', url: addresses_query_url })
+    %div.col-2.offset-1
+      = f.input_field(:housenumber)
+  = f.labeled_input_field(:postbox)
+  = f.labeled_input_field(:zip_code)
+  = f.labeled_input_field(:town)
+
 - if entry.show_attr?(:country)
   = f.labeled(:country) do
     .shown.ms-0


app/views/event/participation_contact_datas/_address_fields.html.haml
```diff
@@ -4,14 +4,17 @@
...
+- if entry.show_address
+  = f.labeled(:address, class: 'd-flex') do
+    %div.col-9
+      = f.input_field(:street, data: { provide: 'entity', updater: 'AddressTypeahead.update', url: addresses_query_url }) if entry.show_attr?(:street)
+    %div.col-2.offset-1
+      = f.input_field(:housenumber) if entry.show_attr?(:housenumber)
+  = f.labeled_input_field(:postbox) if entry.show_attr?(:postbox)
+  = f.labeled_input_field(:zip_code) if entry.show_attr?(:zip_code)
+  = f.labeled_input_field(:town) if entry.show_attr?(:zip_code)
+
```
njaeggi, line 17:
```suggestion
= f.labeled(:address, class: 'd-flex') do
  %div.col-9
    = f.input_field(:street, data: { provide: 'entity', updater: 'AddressTypeahead.update', url: addresses_query_url }) if entry.show_attr?(:street)
  %div.col-2.offset-1
    = f.input_field(:housenumber) if entry.show_attr?(:housenumber)
= f.labeled_input_field(:postbox) if entry.show_attr?(:postbox)
= f.labeled_input_field(:zip_code) if entry.show_attr?(:zip_code)
= f.labeled_input_field(:town) if entry.show_attr?(:town)
```

My suggestion is the old view code, there is just an if statement missing for postbox, I think when you add that, it will solve the bug like I already stated in https://github.com/hitobito/hitobito/pull/3415#pullrequestreview-2867558316

Also with your address changes it would disable ALL address related fields, when a user just marks postbox as hidden, no other address attr

.rubocop_todo.yml
```diff
@@ -125,7 +125,7 @@ RSpec/ImplicitExpect:
     - 'spec/support/crud_controller_test_helper.rb'
     - 'spec/views/contactable/_fields.html.haml_spec.rb'
     - 'spec/views/event/participations/_actions_show.html.haml_spec.rb'
-    - 'spec/views/event/participations/_form.html.haml_spec.rb'
+    - 'spec/views/event/participations/_address_fields.html.haml_spec.rb'
```
njaeggi, line 128: Why? Did you commit this accidentally?

app/models/event/participation_contact_data.rb
njaeggi, line 1: You should be able to revert all those changes with the simple one line solution I described here: https://github.com/hitobito/hitobito/pull/3415#pullrequestreview-2867558316

njaeggi (changes requested):
A simple `if entry.show_attr?(:postbox)` should do the job here, like we do it for all other attrs that work

app/views/event/participation_contact_datas/_address_fields.html.haml
Vakmeth: It would solve the bug but it wouldn't be clean would it? So we do all the checks for all the fields everytime instead of looking initially if we even want to show the address. Also, with this statement we can hide the title "Adresse" which is currently displayed if we only solve the bug by adding the one line condition.

No it wouldn't. If I only mark the postbox as hidden, the other address attrs are displayed as well that's the hack on the method. With this approach we don't check for each attribute if it's needed when we know from beginning that we don't want to display any of them.

njaeggi: True, the address label is/was another bug, nice catch! I corrected your methods to work how they are supposed to work, only street and housenumber are behind the address label, not postbox, zipcode and town

njaeggi:
@Vakmeth I have corrected your show_address method to actually only work with address, thanks for finding and fixing that bug, I also added an extra test case for it, the only last thing that is missing is an actual spec to check that depending on what the hidden option is, if the input field is rendered. 

Please add a spec that loops through all Event::ParticipationContactData.contact_attrs, and check that when it's hidden the input field is not there, and when the contact_attr is not hidden, the input fields is there, so two specs per loop, this will fix future issues with accidentally forgetting the if statement like it happened with postbox. Do that in a new views spec for _fields.html.haml

app/models/event/participation_contact_data.rb
```diff
@@ -7,14 +7,15 @@ class Event::ParticipationContactData
   attr_reader :person
 
   class_attribute :mandatory_contact_attrs,
+    :address_attrs,
```
njaeggi: We don't need this

spec/views/event/participation_contact_datas/_address_fields.html.haml_spec.rb
```diff
@@ -0,0 +1,39 @@
...
+    expect(rendered).to have_text("Adresse")
+  end
+
+  it "should not render address label if no address attributes are set to display" do
+    event.update!(hidden_contact_attrs: [:housenumber, :street])
+    render locals: {entry: participation_contact_data(event), event:, group:}
+    expect(rendered).to have_no_text("Adresse")
+  end
+
+  def participation_contact_data(event)
```
njaeggi: Like I already said in: https://github.com/hitobito/hitobito/pull/3415#discussion_r2111705106

Please put this inside a let, I will do it myself now, but just for next time, don't resolve any comments without changing or commenting it.

njaeggi (changes requested):

spec/views/event/participation_contact_datas/_address_fields.html.haml_spec.rb
Vakmeth: I did that on purpose because in another test with the same use case @mtnstar fixed this the same way, with a method outside the spec. For reasons of consistency I decided to to this the same way.

njaeggi: Yes, but in his usecase it made more sense, since he passed attributes as a parameter, which were different in each spec, you just passed event as a parameter, that always stayed the same.

Consistency is nice, but in that case the other code was for a different use case

PR #3419, opened 2025-05-26
Return false for group_read_in_this_group? when group is not set

njaeggi:
Sentry Issue: https://sentry.puzzle.ch/pitc/hitobito-backend/issues/75718/

GLP checks this method in some extra call for `donor_visible?` expecting the group to be set, the search is global tho and not group based so in the case of the group not being set we return false

### DIFF
--- a/app/abilities/person_readables.rb
+++ b/app/abilities/person_readables.rb
@@ -92,6 +92,8 @@ class PersonReadables < GroupBasedReadables
   end
 
   def group_read_in_this_group?
+    return false unless group
+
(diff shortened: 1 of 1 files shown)


amaierhofer (dismissed):
Wäre der check nicht viel besser  im glp wagon (hitobito_glp/app/decorators/glp/person_decorator.rb) aufgehoben?

njaeggi:
Replaced with: https://github.com/hitobito/hitobito_glp/pull/74

PR #3437, opened 2025-06-06
Integrate patches with statuscope (#3298)

amaierhofer:


kronn (approved):
This is better as it highlights errors better. But I think in general it would be better to separate this out into a dedicated workflow that is triggered on push.

PR #3446, opened 2025-06-12
Automate running all wagon tests on core changes #3442

carlobeltrame:
Before merging, this needs slight modifications to the tests.yml workflows in each wagon, i.e. add a distinct_id parameter and pass it to the reusable workflow, and also accept the wagon_dependency_ref parameter, even if it is unused.
- [x] [youth](https://github.com/hitobito/hitobito_youth/commit/5ee3927560c890e169597e6e0f3e39334d77a638)
- [x] [pbs](https://github.com/hitobito/hitobito_pbs/commit/62d49bc90bfb0bb2b4d262aa809700c0f8bbe517)
- [x] all other wagons
- [x] Remove the nightly trigger for running all wagon tests (have to do this in all wagons too)

To test this new workflow locally, I used the [act](https://nektosact.com/) plugin for github CLI:
```bash
# One-time setup
gh extension install https://github.com/nektos/gh-act
docker pull catthehacker/ubuntu:full-latest # caution: takes up a lot of disk space!

# Run this to test the workflow
gh act push -P ubuntu-latest=catthehacker/ubuntu:full-latest --pull=false -W .github/workflows/run-all-wagon-tests.yml -s WAGON_TESTS_GITHUB_TOKEN=github_pat_<redacted>
```

### DIFF
--- a/.github/workflows/run-all-wagon-tests.yml
+++ b/.github/workflows/run-all-wagon-tests.yml
@@ -1,0 +1,86 @@
+name: 'Run the tests in all wagons with a specific core revision'
+
+on:
+  workflow_dispatch:
+    inputs:
+      core_ref:
+        description: Use a specific version of the core for the workflow run. Defaults to master.
+        default: ""
+      wagon_dependency_ref:
+        description: Use a specific version of the wagon dependency for the workflow run. Defaults to master.
+        default: ""
+      wagon_ref:
+        description: Use a specific version of the wagon dependency for the workflow run. Defaults to master.
+        default: "master"
+  push:
+    branches: [ master ]
+    paths-ignore:
+      - 'doc/**'
+      - '**.md'
+  pull_request:
+    paths-ignore:
+      - 'doc/**'
+      - '**.md'
+  secrets:
+    WAGON_TESTS_GITHUB_TOKEN:
+      description: 'Token with repo and actions:read and actions:write for running the tests in all wagon repos'
+      required: true
+
+jobs:
+  find-wagons:
+    name: 'Find all Puzzle-managed wagons'
+    runs-on: 'ubuntu-latest'
+    outputs:
+      wagons: ${{ steps.list-wagon-repositories.outputs.wagons }}
+
+    steps:
+      - name: 'List wagon repositories'
+        id: list-wagon-repositories
+        run: |
+          WAGONS=$(gh repo list hitobito -L 100 --no-archived --json owner,name | \
+            jq 'map(select(.name | test("^hitobito_"))) | map({ owner: .owner.login, repo: .name, name: .name | match("^hitobito_(.+)$").captures[0].string })')
+          echo "wagons=$WAGONS" | tr -d "\n" >> $GITHUB_OUTPUT
+        env:
+          GH_TOKEN: ${{ secrets.WAGON_TESTS_GITHUB_TOKEN }}
+
+  run-wagon-tests:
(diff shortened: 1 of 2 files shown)


.github/workflows/run-all-wagon-tests.yml
```diff
@@ -0,0 +1,86 @@
...
+        description: Use a specific version of the wagon dependency for the workflow run. Defaults to master.
+        default: ""
+      wagon_ref:
+        description: Use a specific version of the wagon dependency for the workflow run. Defaults to master.
+        default: "master"
+  push:
+    branches: [ master ]
+    paths-ignore:
+      - 'doc/**'
+      - '**.md'
```
kronn, line 19: I am unsure about the `push` trigger.

I suggest to at least skip runs where only the VERSION or the CHANGELOG change. 
But: Can we maybe omit the push-trigger completely?

Alternatively, if we keep it, we could remove the schedule in the wagons. If every push to the core triggers all the wagons, the integration-test does not need to be checked on a schedule, but becomes event-driven. Since the remote workflows all notify statuscope, the dashboard continues to work as expected.

.github/workflows/run-all-wagon-tests.yml
```diff
@@ -0,0 +1,86 @@
...
+    outputs:
+      wagons: ${{ steps.list-wagon-repositories.outputs.wagons }}
+
+    steps:
+      - name: 'List wagon repositories'
+        id: list-wagon-repositories
+        run: |
+          WAGONS=$(gh repo list hitobito -L 100 --no-archived --json owner,name | \
+            jq 'map(select(.name | test("^hitobito_"))) | map({ owner: .owner.login, repo: .name, name: .name | match("^hitobito_(.+)$").captures[0].string })')
+          echo "wagons=$WAGONS" | tr -d "\n" >> $GITHUB_OUTPUT
```
carlobeltrame, line 45: For easier debuggability and for new devs trying to understand what's going on, maybe echo the result normally as well

.github/workflows/run-all-wagon-tests.yml
```diff
@@ -0,0 +1,86 @@
...
+    name: 'Run ${{ matrix.wagon.name }} wagon tests'
+    runs-on: 'ubuntu-latest'
+    needs:
+      - find-wagons
+    strategy:
+      fail-fast: false
+      matrix:
+        wagon: ${{ fromJSON(needs.find-wagons.outputs.wagons) }}
+    steps:
+      - run: echo $JSON
```
carlobeltrame: Remove debug output or save the json into a variable that multiple steps can use, to make sure what's output here matches what's sent to the wagon workflow

carlobeltrame (commented):
Todo: Currently, the access token is a personal access token generated from my own account. It will expire on 2026-06-12 because github PAT cannot be valid for longer than 1 year. Do we have or want a service account to generate this token, so it's not tied to a dev's account?

.github/workflows/run-all-wagon-tests.yml
kronn: verbose output of non-confidentials parts is a good thing.

.github/workflows/run-all-wagon-tests.yml
carlobeltrame: Outcome of discussion at daily:
- Keep the push trigger, wagon tests should be a normal part of the core CI from now on
- Ignore VERSION. CHANGELOG is already a .md file and is already ignored
- Remove the nightly triggers on all wagon repos

PR #3480, opened 2025-06-27
Run the wagon tests directly in the core

carlobeltrame:
Not triggering and awaiting them in the remote repository means fewer job runners just idling to wait for some other job to complete.

~~To be solved is the detection of the wagon dependency repository. Also, this change is thus far untested.~~

### DIFF
--- a/.github/workflows/run-all-wagon-tests.yml
+++ b/.github/workflows/run-all-wagon-tests.yml
@@ -52,7 +52,6 @@ jobs:
 
   run-wagon-tests:
     name: 'Run ${{ matrix.wagon.name }} wagon tests'
-    runs-on: 'ubuntu-latest'
     needs:
       - find-wagons
     strategy:
@@ -60,27 +59,12 @@ jobs:
       matrix:
         wagon: ${{ fromJSON(needs.find-wagons.outputs.wagons) }}
       max-parallel: 5 # needed so we don't occupy all the runners with core waiting jobs and the wagon tests can never start
-    # steps:
-    #   - name: "Determine common branch and set branch name for wagon"
-    #     env:
-    #       INPUT_WAGON_REF: ${{ inputs.wagon_ref }}
-    #     run: |
-    #       COMMON_BRANCH_NAME="${{github.head_ref || github.ref_name}}"
-    #       if [ -z "$INPUT_WAGON_REF" ]; then
-    #         WAGON_REF=$(git ls-remote --heads https://github.com/${{matrix.wagon.owner}}/${{matrix.wagon.repo}} | cut -d/ -f3- | grep -x "$COMMON_BRANCH_NAME" || echo 'master')
-    #       fi
-    #
-    #       echo "WAGON_REF=$WAGON_REF" >> $GITHUB_ENV
-    #     shell: bash
-    #
-    #  core_ref: env.WAGON_REF
-    #  wagon_dependency_ref: env.WAGON_REF
 
     uses: hitobito/hitobito/.github/workflows/wagon-tests.yml@master
     with:
       wagon_repository: ${{ github.event.repository.name }}
-      core_ref: ${{ github.ref || inputs.core_ref || 'sac-master' }}
-      wagon_dependency_ref: ${{ inputs.wagon_dependency_ref || 'sac-master' }}
+      core_ref: ${{ github.ref || inputs.core_ref || '' }}
+      wagon_dependency_ref: ${{ inputs.wagon_dependency_ref || '' }}
     secrets:
       HEARTBEAT_URL: ${{ secrets.HEARTBEAT_URL }}
       HEARTBEAT_TOKEN: ${{ secrets.HEARTBEAT_TOKEN }}
--- a/.github/workflows/run-all-wagon-tests.yml
+++ b/.github/workflows/run-all-wagon-tests.yml
@@ -43,7 +43,7 @@ jobs:
         id: list-wagon-repositories
         run: |
           WAGONS=$(gh repo list hitobito -L 100 --no-archived --json owner,name | \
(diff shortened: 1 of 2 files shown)


.github/workflows/wagon-tests.yml
```diff
@@ -41,6 +41,9 @@ defaults:
     working-directory: hitobito
 
 jobs:
+  detect_wagon_dependency:
+    # TODO
```
kronn: Vorschlag:
- checkout/cd wagon-repository
- `grep "add_dependency.*hitobito_" *.gemspec | sed 's/.*hitobito_\([-_a-z]*\).*/\1/'`

Das gibt z.B. für die Jubla "youth" zurück. man kann das sed sicherlich noch anpassen, dass man direkt "hitobito_youth" zurückbekommt.

carlobeltrame: Sounds good, I was thinking along the same lines. Maybe the git checkout could be sparse (only checkout the gemspec or lib dir) and only checkout the master branch.

kronn: the gemspec should be enough. doesn't need to be a checkout, a curl would be enough, like this: 

https://github.com/hitobito/hitobito/blob/f59bf88b01c3cfe8087739254cd26f9492c7a5b8/.github/workflows/next-version.yml#L54-L60

This gets the version-script. but if we know the wagon-name, we could only get the gemspec. The extraction from repo-name to project-identifier could be copied from here:

https://github.com/hitobito/hitobito/blob/f59bf88b01c3cfe8087739254cd26f9492c7a5b8/.github/workflows/stage-settings.yml#L89-L90

.github/workflows/run-all-wagon-tests.yml
```diff
@@ -59,39 +59,11 @@ jobs:
...
-              "wagon_dependency_ref": "${{ inputs.wagon_dependency_ref || 'sac-master' }}"
-            }
-
-      - name: Link to the triggered wagon tests run
-        run: 'echo "${{ matrix.wagon.name }} tests triggered: ${{ steps.run-wagon-tests.outputs.run_url }}"'
-
-      - name: Wait for ${{ matrix.wagon.name }} wagon tests result
-        uses: carlobeltrame/await-remote-run@conditional-api-requests
-        with:
-          token: ${{ secrets.WAGON_TESTS_GITHUB_TOKEN }}
```
carlobeltrame, line 106: With this change, we can also just use the default github token on line 50, no need anymore to use a PAT with write access

carlobeltrame:
Statuscope integration needs to be rethought with this approach, assuming each wagon has a dedicated statuscope URL to post to, which cannot easily be deduced from just the wagon name, right?

kronn:
for changes in the core which then fail the wagon, we could trigger a statuscope-workflow in the wagon. This would mean separating it into a workflow that solely exists to be triggered in the context of the wagon-repo.

This would result in secrets being read from the corresponding repo. The secrets then contains the token and identifier for statusscope.

The alternative would be to let said tokens live in two places. Once in the wagon for those working on features there. And once, with a slightly different naming-scheme in the core-settings to allow for fine-grained information about failures.

A third option would be to have one "wagon-tests"-identifier in statuscope that simply informs us THAT something is broken. This is better than nothing, but clearly worse than fine-grained information. OTOH, it would nicely inform WHERE the failure-info can be found.

.github/workflows/wagon-tests.yml
```diff
@@ -311,15 +368,15 @@ jobs:
 
   notify_statuscope:
     uses: ./.github/workflows/notify-statuscope.yml
-    needs: [wagon_rubocop, wagon_specs, wagon_feature_specs, wagon_seeds, wagon_pending_specs]
+    needs: [detect_wagon_dependency, wagon_rubocop, wagon_specs, wagon_feature_specs, wagon_seeds, wagon_pending_specs]
```
kronn, line 371: I suggest removing this as it is an implied and for this notification unneeded dependency. 

```suggestion
    needs: [wagon_rubocop, wagon_specs, wagon_feature_specs, wagon_seeds, wagon_pending_specs]
```

.github/workflows/wagon-tests.yml
```diff
@@ -311,15 +368,15 @@ jobs:
...
     if: >-
       ( success() || failure() ) &&
       ( github.ref_name == 'master' || github.ref_name == 'stable' ) &&
       ( contains((inputs.core_ref || inputs.main_branch), inputs.main_branch) ) &&
       ( contains((inputs.wagon_dependency_ref || inputs.main_branch), inputs.main_branch) )
     with:
       repository: ${{ inputs.wagon_repository }}
-      test_result: ${{ needs.wagon_rubocop.result == 'success' && needs.wagon_specs.result == 'success' && needs.wagon_feature_specs.result == 'success' && needs.wagon_seeds.result == 'success' && needs.wagon_pending_specs.result == 'success'}}
+      test_result: ${{ needs.detect_wagon_dependency.result == 'success' && needs.wagon_rubocop.result == 'success' && needs.wagon_specs.result == 'success' && needs.wagon_feature_specs.result == 'success' && needs.wagon_seeds.result == 'success' && needs.wagon_pending_specs.result == 'success'}}
```
kronn, line 379: I think the result of the dependency-check is irrelevant. if that would fail, but the specs are still green, it is still a success IMHO. Therefore I suggest this:


```suggestion
      test_result: ${{ needs.wagon_rubocop.result == 'success' && needs.wagon_specs.result == 'success' && needs.wagon_feature_specs.result == 'success' && needs.wagon_seeds.result == 'success' && needs.wagon_pending_specs.result == 'success'}}
```

kronn (changes requested):
I am unsure if every jobs needs the dependency-wagon. for the statuscope-notification, I suggest removing it. For the other I simply want to question it. if the consistency is a benefit, keep it. those who do not need the dependency-wagon are running small and fast jobs anyway, so there is little benefit in them being started earlier.

PR #3523, opened 2025-07-11
Activate nil check when handling bounces

carlobeltrame:
Not sure this is the right fix, and not sure why this line was commented before.
https://sentry.puzzle.ch/pitc/hitobito-backend/issues/77643/

Tritt bisher nur auf Multitenancy auf.

### DIFF
--- a/app/utils/imap/bounce_mail.rb
+++ b/app/utils/imap/bounce_mail.rb
@@ -55,6 +55,7 @@ module Imap
     def recipients_of_original_message
       undelivered_message = mail.parts.find { |part| part.content_description == "Undelivered Message" }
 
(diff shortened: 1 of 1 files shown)


app/utils/imap/bounce_mail.rb
```diff
@@ -55,7 +55,7 @@ def bounced_mail_addresses
     def recipients_of_original_message
       undelivered_message = mail.parts.find { |part| part.content_description == "Undelivered Message" }
 
-      # return [] if undelivered_message.blank?
+      return [] if undelivered_message.blank?
```
kronn, line 59:
```suggestion
      # this mostly happens when there is no (parseable) undelivered messages attached
      return [] if undelivered_message.blank?
```

PR #3539, opened 2025-07-17
Handle both arrays and hashes for the param

kronn:
https://sentry.puzzle.ch/pitc/hitobito-backend/issues/77225/


daniel-illi (approved):
LGTM :+1:

PR #3625, opened 2025-09-15
Allow wagons to define title for contact on custom events show

philippks:
As needed for fundraisings, see https://github.com/hitobito/hitobito_dsj/issues/34#issuecomment-3233101810


app/views/events/_attrs_contact.html.haml
```diff
@@ -1,4 +1,9 @@
+-#  Copyright (c) 2012-2025, Jungwacht Blauring Schweiz. This file is part of
```
carlobeltrame:
```suggestion
-#  Copyright (c) 2012-2025, Dachverband Schweizer Jugendparlamente. This file is part of
```

PR #3722, opened 2025-11-06
Retry the release when it fails due to race conditions

carlobeltrame:
In the case of pushing the translations and version tag to the core, the `git pull --rebase` before `git push` was not enough.
See failures [here](https://github.com/hitobito/hitobito/actions/runs/19134803436/job/54684120288) and [here](https://github.com/hitobito/hitobito/actions/runs/19132898502/attempts/1).

To address this, we can retry on failure.

### DIFF
--- a/lib/release/tooling.rb
+++ b/lib/release/tooling.rb
@@ -38,7 +38,7 @@ module Release
         end
 
         notify "Attempt #{attempt} failed: #{e.message}. Resetting to upstream and retrying..."
-        execute "git reset --hard @{u}"
+        execute "git reset --hard @{u}" # @{u} is a special git ref for 'current-upstream-branch'
 
         sleep_time = 2**attempt
         notify "Waiting #{sleep_time} seconds before retry..."


lib/release/lowlevel.rb
```diff
@@ -47,14 +47,39 @@ def submodule_status
...
+      begin
+        yield
+      rescue => e
+        if attempt >= max_attempts
+          notify "Failed after #{max_attempts} attempts: #{e.message}"
+          raise e
+        end
+
+        notify "Attempt #{attempt} failed: #{e.message}. Resetting to upstream and retrying..."
+        execute "git reset --hard @{u}"
```
kronn: Where does this `@{u}` come from? Should this be "upstream"?

lib/release/lowlevel.rb
```diff
@@ -47,14 +47,39 @@ def submodule_status
       execute "git submodule status"
     end
 
+    # rubocop:disable Metrics/MethodLength
+    def retry_with_reset(max_attempts: 5, &block)
```
kronn: I would put this method in the Tooling-module. I know the distinction is mostly made-up, but still.

lib/release/lowlevel.rb
carlobeltrame: `@{u}` (not to be confused with ruby string interpolation `#{u}`) is a feature of git itself. It's a special ref that refers to the current branch, but on the origin (well, to be precise, it's the origin branch tracked by the current local branch).

https://stackoverflow.com/a/61961340

Since the same `retry_with_reset` method is used on the core, wagons and composition repo, and the target_branch may also vary, I don't necessarily know the current branch name.

So this line essentially says: Reset my local state to the way they are on the tracked upstream branch; or: discard anything in my local clone that isn't contained in the origin's version of the connected branch.

Does this need some kind of explanatory comment?

kronn: Yes, something like "@{u} is git-speak for 'current-upstream-branch'" should give enough information.

kronn (approved):

PR #3761, opened 2025-11-20
Clear remember_me on password reset (hitobito/hitobito_sww#310)

philippks:
An error ocurred when the password of a logged in user (with remember_me) was resetted and the user opened a page with the old cookie.

An further improvement would be, to properly handle the bad state in the controller.

For now, we just try to avoid the bad state and cleanup existing data:

```
Person.where(encrypted_password: nil).where.not(remember_created_at: nil).update_all remember_created_at: nil
```

### DIFF
--- a/app/controllers/person/security_tools_controller.rb
+++ b/app/controllers/person/security_tools_controller.rb
@@ -41,6 +41,7 @@ class Person::SecurityToolsController < ApplicationController
 
   def password_override
     person.encrypted_password = nil
+    person.forget_me!
     person.save
     notify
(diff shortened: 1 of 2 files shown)


carlobeltrame (approved):
Looks great! How about blocking the person on the security tools view, do we need something similar there? Or is the block applied after the login?

philippks:
> Looks great! How about blocking the person on the security tools view, do we need something similar there? Or is the block applied after the login?

Tested locally,  `blocked_at` is checked after the login: 
image

PR #3855, opened 2025-12-29
Cleanup pending specs, enable ci check

amaierhofer:


app/views/roles/_modal.html.haml
```diff
@@ -4,7 +4,9 @@
 -#  https://github.com/hitobito/hitobito.
 
 - action = entry.persisted? ? :update : :create
-= standard_form [group, entry], controller: :roles, action: action, remote: true  do |f|
+- action_path = entry.persisted? ? [@original_group || group, entry] : [group, entry]
```
kronn, line 7: not a fan of code-golfing in prod. Also, the conditional is very similar. maybe this can be written better, but I do not see it yet. 

For now, it is okay...

app/views/roles/_popover.js.haml
```diff
@@ -6,10 +6,17 @@
...
+
+  if (instance) {
+    instance.dispose();
+  }
+
+  Popover.getOrCreateInstance(el, {
+    container: "body",
+    sanitize: false,
+    html: true,
+    placement: "auto",
```
kronn: why the change from `bottom` to `auto`?

kronn (approved):

app/views/roles/_popover.js.haml
amaierhofer: I changed that back to `bottom`

app/views/roles/_modal.html.haml
amaierhofer: Agreed, this is due to the complexicity of changing role (and possible group of role), would probably deserve a seperate view/controller, but lets go with that small fix for now

PR #3864, opened 2026-01-07
Move person/invoices sheet and views to contactables/invoices

njaeggi:
I noticed that in https://github.com/hitobito/hitobito/commit/edcd6416674059e619d2bd2e4b3c23c80f227c7d the view page for received_invoices broke, this should fix it :smile:


carlobeltrame (approved):
Tested, works. Thanks!

PR #3874, opened 2026-01-07
Use InvoiceController as parent class for Contactable::InvoiceController

njaeggi:
https://github.com/hitobito/hitobito/commit/9ad31fad192876e42cc62d067fb168205ca1a12a broke the view page for invoices on person (and groups in the future) because filter_params is not defined in the `contactable/invoices_controller`. I fixed it by just using the `InvoiceController` as a parent class instead of directly using the `ListController`. This fixes the issue and also makes much more sense in my Opinion, instead of completly seperating the invoice controllers, maybe it has a specific reason why, in that case we would have to define the filter_params in the controller seperatly.

### DIFF
--- a/app/controllers/contactables/invoices_controller.rb
+++ b/app/controllers/contactables/invoices_controller.rb
@@ -3,7 +3,7 @@
 #  or later. See the COPYING file at the top-level directory or at
 #  https://github.com/hitobito/hitobito.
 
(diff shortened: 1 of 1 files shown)


carlobeltrame:
Already fixed in b96a457a4e2d06b1421f9ff15fd7d21daca7ad4c

PR #3905, opened 2026-01-23
adjust tabbing order

RandomTannenbaum:


app/javascript/stylesheets/hitobito/_layout.scss
```diff
@@ -82,8 +88,9 @@ body {
     }
     width: 30%;
     max-width: 25rem;
+    height: 100%;
     z-index: 1;
-    margin-top: $logo-height + 40px;
+    margin-top: calc($logo-height + 2px - 1rem);
```
RandomTannenbaum, line 93: The old `margin-top` value of the sidenav was `$logo-height + 40px`. This margin used to reference the top of the main content, which also was where the quicksearch was located. Now the quicksearch is in the header above the main content, which means that the margin of the sidenav now references the bottom of that header. The quicksearch has a height of `38px` and the gap between the header and the main content is `1rem`. -> `$logo-height + 40px - 38px - 1rem` = `$logo-height + 2px - 1 rem`.

Maybe we can find a better solution for this as this would break as soon as the header height changes.

app/views/layouts/application.html.haml
```diff
@@ -34,62 +34,67 @@
...
-              .btn-toolbar= yield :toolbar
-
-            -# div must always exist to insert flash messages with ajax
-            #flash
-              = render partial: 'layouts/flash', collection: [:notice, :alert], as: :level
-
-            = content_for?(:content) ? yield(:content) : yield
+      %header.main-header
+        - if render_nav?
+          .nav-left
```
kronn, line 43: this seems unnecessary. can we can rid of this empty div?

spec/regressions/qualifications_controller_spec.rb
```diff
@@ -53,7 +53,7 @@ def it_should_redirect_to_index
         expect(sheet.find_link("Personen")[:href]).to eq group_people_path(top_group, returning: true)
         expect(sheet.find_link("Top Leader")[:href]).to eq group_person_path(top_group, top_leader)
 
-        nav = page.find(".nav-left")
+        nav = page.find("nav.nav-left")
```
kronn, line 56: maybe this is not needed when the empty `<div class="nav-left">` is removed...

kronn (changes requested):
I think a little cleanup is needed.

app/views/layouts/application.html.haml
RandomTannenbaum: Since the quicksearch is now above the sidebar and main content it would be aligned with the left side of the screen. Previously the search was in the main content and thus aligned with the left border of it. The empty div acts as a spacer so the search bar is, like before, aligned with the main content, which is located right beside the actual nav. This approach has the advantage of also working when the nav-left style has been overwritten in a wagon.

PR #3909, opened 2026-01-26
Migrate all invoice_run recipient options to recipient_source (#3860)

njaeggi:


config/locales/models.de.yml
```diff
@@ -506,6 +506,7 @@ de:
         origin: Host mit API-Zugriff
 
       person:
+        id: Personennummer
```
njaeggi, line 509: We use "Personennummer" as a translation here, since that will be less confusing for users than getting "Id" inside the people filter field selection.

app/views/invoices/_form.html.haml
```diff
@@ -5,6 +5,7 @@
 
 = entry_form(data: { group: group_path(parent) }, class: 'form') do |f|
   = f.hidden_field :recipient_id
+  = f.hidden_field :recipient_type
```
njaeggi, line 8: This is needed to set the recipient_type and prevent the bug that single invoices don't have a type yet.
https://chat.puzzle.ch/channel/hitobito-office?msg=fuLkSG2GtKDS63Knj

db/migrate/20260127231959_set_all_invoice_recipient_type_to_person.rb
njaeggi, line 1: https://chat.puzzle.ch/channel/hitobito-office?msg=fuLkSG2GtKDS63Knj

njaeggi:
All Wagon Spec runs with youth wagon:
- https://github.com/hitobito/hitobito_sjas/actions/runs/21472140415
- https://github.com/hitobito/hitobito_sac_cas/actions/runs/21475533622

kronn (approved):

PR #3910, opened 2026-01-26
 Use first predefined_label in export for unlabeled phone_numbers 

njaeggi:
It is not possible to export participations in events with guests, since guests only have a phone_number as a string. This leads to an error during the export because there is no phone_number label.

In the case of guests (or if the label would for some other reason be nil) we want to use the first option of predefined_labels from `settings.yml`.

### DIFF
--- a/app/domain/export/tabular/people/contact_accounts.rb
+++ b/app/domain/export/tabular/people/contact_accounts.rb
@@ -7,17 +7,17 @@ module Export::Tabular::People
   module ContactAccounts
     class << self
       def key(model, label)
-        :"#{model.model_name.to_s.underscore}_#{(label || default_label(model)).downcase}"
+        :"#{model.model_name.to_s.underscore}_#{label_or_default(label, model).downcase}"
       end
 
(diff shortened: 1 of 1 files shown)


app/domain/export/tabular/people/contact_accounts.rb
```diff
@@ -7,11 +7,17 @@ module Export::Tabular::People
   module ContactAccounts
     class << self
       def key(model, label)
-        :"#{model.model_name.to_s.underscore}_#{label.downcase}"
+        :"#{model.model_name.to_s.underscore}_#{(label || default_label(model)).downcase}"
```
carlobeltrame: Nitpick: Ich fände es schöner, wenn hier nicht viel Logik in der String Interpolation drin wäre. Kannst du das `label ||` und das `.downcase` auch noch in die Methode reinverschieben, und die Methode z.B. `label_or_default(label, model)` oder `label_fallback(label, model)` oder sonst wie nennen?

PR #3938, opened 2026-02-09
bugfix/improve version decorator

daniel-illi:
- **Improve rendering of papertrail create versions**

### DIFF
--- a/config/locales/views.de.yml
+++ b/config/locales/views.de.yml
@@ -2298,7 +2298,7 @@ de:
     household_destroyed: "Haushalt (%{object_changes}) wurde aufgelöst."
     person_merge: "Zwei Personen wurden in diesen Eintrag zusammengeführt. Gelöschte Person:"
     deleted_service_token: "Inzwischen gelöschtes %{model_name}"
-    deleted_user: "Inzwischen gelöschte Person %{id}"
+    deleted_user: "inzwischen gelöschte Person %{id}"
 
   tags/merge:
     success: Die Tags wurden erfolgreich zusammengeführt.


app/decorators/paper_trail/version_decorator.rb
```diff
@@ -37,33 +37,38 @@ def link_to_changed_object(path)
...
-        else
-          author_person(author_id)
-        end
+      return if model.version_author.blank?
+
+      case model.whodunnit_type
+      when ServiceToken.sti_name
+        author_service_token || deleted_service_token_message
+      when Person.sti_name
+        author_person || deleted_user_message
```
kronn, line 46: this also shows the `deleted_user_message` if the user cannot be displayed due to permission errors. Then the error-message might be misleading. This is similar to the handling of the token. Maybe still an improvement, but also not perfectly worded.

config/locales/views.de.yml
```diff
@@ -2290,6 +2290,7 @@ de:
     household_destroyed: "Haushalt (%{object_changes}) wurde aufgelöst."
     person_merge: "Zwei Personen wurden in diesen Eintrag zusammengeführt. Gelöschte Person:"
     deleted_service_token: "Inzwischen gelöschtes %{model_name}"
+    deleted_user: "inzwischen gelöschte Person %{id}"
```
kronn, line 2301:
```suggestion
    deleted_user: "Inzwischen gelöschte Person %{id}"
```

kronn (changes requested):

app/decorators/paper_trail/version_decorator.rb
daniel-illi: are you sure about this? :thinking: 
As I understand the code, it links to the Person if the user `can?(:read)` the Person, otherwise it shows the `Person#to_s` 
This is not new behaviour, just restructured. And there is also a spec for this (line 80 in `version_decorator_spec.rb`).

PR #3951, opened 2026-02-17
Parse date value in datetime setter (hitobito/hitobito_sac_cas#2262)

njaeggi:
This pull request solves an issue of not being able to track actual changes to event dates date fields (or other models including `DatetimeAttribute`).

The setter was comparing a string to a date which basically led to `start_at_will_change!` and `finish_at_will_change!` always triggering. This added start_at and finish_at to `changes`, when nothing has changed, which then breaks all the methods for active record dirty tracking:
- `will_save_change_to_start_at?`
- `changes_to_save`
- `changed?`
- `saved_change_to_start_at?`
- and many more: https://api.rubyonrails.org/classes/ActiveRecord/AttributeMethods/Dirty.html

### DIFF
--- a/app/models/concerns/datetime_attribute.rb
+++ b/app/models/concerns/datetime_attribute.rb
@@ -58,6 +58,20 @@ module DatetimeAttribute
     @datetimes[attr] ||= {}
   end
 
+  def normalize_datetime_field(field, value)
+    (field == :date) ? parse_date(value) : value.to_i
+  end
+
+  def parse_date(value)
+    if value.split(".").last.length == 2
+      Date.strptime(value, "%d.%m.%y")
+    else
+      Date.parse(value)
+    end
+  rescue
+    value
+  end
+
   module ClassMethods
     def datetime_attr(*attrs)
       attrs.each do |attr|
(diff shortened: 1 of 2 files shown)


app/models/concerns/datetime_attribute.rb
```diff
@@ -58,6 +58,23 @@ def datetime_fields(attr)
     @datetimes[attr] ||= {}
   end
 
+  def normalize_datetime_field(field, value)
+    case field
+    when :date then parse_date(value)
+    when :hour, :min then value.presence&.to_i
```
amaierhofer: Kannst du prägnater schreiben, das `presence` braucht es hier nicht, `nil.to_i` liefert ebenfalls 0 und der `else` fall kann nicht eintreten.

```
def normalize_datetime_field(field, value)
  (field == :date) ? parse_date(value) : value.to_i
end
```

app/models/concerns/datetime_attribute.rb
```diff
@@ -58,6 +58,23 @@ def datetime_fields(attr)
...
+    else value
+    end
+  end
+
+  def parse_date(value)
```
amaierhofer, line 65: Kannst du prägnanter schreiben

```
def parse_date(value)
  Date.parse(value)
rescue
  value
end
````

app/models/concerns/datetime_attribute.rb
amaierhofer: Das Verhalten, dass `nil` oder ungültige String values 0 zurückliefern würde noch eine spec verdienen.

amaierhofer (changes requested):

PR #3957, opened 2026-02-19
Recipient group type UI

### LINKED ISSUE #3749: INVOICE: Auswahl des Empfänger-Gruppentyps
# Ausgangslage
Bei SWB wird via settings.yml die Wahl der Gruppentypen global definiert.
In #3860 wird/wurde die Möglichkeit geschaffen, mit einer Parent-Gruppe und einem Gruppentyp eine gefilterte Liste von Gruppen in der DB zu spezifizieren (GroupsFilter, ähnlich wie PeopleFilter).

# Anforderungen
Es muss möglich sein, verschiedene Sammelrechnungen (auch von der gleichen Ebene aus) an verschiedene Gruppentypen zu stellen. Z.B. vom Dachverband an Regionen, Vereine oder Teams, und von der Region aus an Vereine oder Teams.
Diese Auswahl von Gruppentyp und Parent-Gruppe soll zu einem GroupsFilter gemacht werden können, damit zukünftige Erweiterungen am GroupsFilter auch bei Sammelrechnungen greifen können.

# Abgrenzungen
- Muss man beim Fahren eines Rechnungslaufs oder beim Definieren des Empfänger-Gruppentyps einzelne Empfängergruppen manuell an- oder abwählen können?
  - Beim SBV ist das aktuell möglich, aber war nie explizit gewünscht in den ursprünglichen Anforderungen (und die Gruppenrechnungen werden beim SBV aktuell nicht produktiv eingesetzt).
  - Bei SWB ist das aktuell nicht möglich.
  - Bei den deutschen Pfadis ist es bisher keine Anforderung, abgesehen von archivierten Gruppen.
  - Beim EJV ist eine Anforderung, dass Gruppen via Flag angeben können, ob der Dachverband der Gruppe eine Rechnung stellen soll oder immer direkt den Mitgliedern. Das könnte man auch via Wagon-Customization an der Berechnung der Empfängergruppen lösen. Dann würde es aber komplett unmöglich, solchen Gruppen eine Sammelrechnung zu stellen, ausser wir ermöglichen, dies bei EJV im UI der Sammelrechnung zu aktivieren oder deaktivieren.
  - Vorschlag: Vorerst mal weglassen und später als separates Feature dazuprogrammieren, wenn es jemand explizit braucht. Dies kann später als Erweiterung vom GroupsFilter gemacht werden.
- Zu Empfänger-Gruppen, welche während oder nach der Rechnungsperiode archiviert oder gelöscht wurden:
  - [name=di] Ich denke da sollte geprüft werden, ob das Archivierungsdatum in der Rechnungsperiode liegt. Je nachdem wäre aber auch argumentierbar dass die Gruppen nur ausgenommen sind wenn das Datum VOR der Rechnungsperiode liegt..
  - [name=pz] Ich denke das archivieren und löschen erfolgt ja oft erst, wenn eine Gruppe schon eine Weile inaktiv war. Ich würde das vorerst ausklammern.
  - [name=tel] Das gruppen archivierungsthema erstmal zu ignorieren und die entscheidung zu vertagen finde ich auch in Ordnung. Bei den deutschen Pfadis sollte es dass zumindest im ersten Jahr nicht geben. Haben wir am Ende noch budget ist es sicherlich gut sie darauf aufmerksam zu machen/selbst zu wissen dass es da limitationen gibt. Notfalls erweitern wir dass dan wenn der erste Kunde probleme hat (bestenfalls haben wir am schluss natürlich noch budget dafür..)

# Offene Fragen

# Mockup

```mermaid
classDiagram
class PeriodInvoiceTemplate {
  ...
  has_one :recipient_source, class: GroupsFilter
}

class GroupsFilter {
  belongs_to :parent
  serialize :group_types
  active_at Date
}

PeriodInvoiceTemplate "0..1" --> "1" GroupsFilter
```

Achtung: Die Auswahl des Gruppentyps kann auch durch ein einfaches select-Dropdown umgesetzt werden, statt durch einen Link mit Dialog wie in diesem Mockup.

# Tech-Spec
- Auf der Sammelrechnung kann ein Gruppentyp ausgewählt werden, der unterhalb der Ebene möglich ist.
- Die Bedeutung dieses Gruppentyps ist _"Alle Gruppen von diesem Typ in oder unterhalb dieses Layers sind Empfänger einer Rechnung. Ausgenommen sind Gruppen, welche bereits beim Startdatum der Rechnungsperiode gelöscht oder archiviert waren."_
- Es ist egal ob in oder unterhalb des Layers eine Gruppe von diesem Typ existiert oder nicht. Alle theoretisch möglichen Gruppentypen sind auswählbar.
- Beim Speichern der Sammelrechnung wird ein GroupsFilter angelegt / der bereits verknüpfte GroupsFilter aktualisiert.
  - parent_id wird auf period_invoice_template.group_id gesetzt
  - group_types wird auf den einen im UI ausgewählten Gruppentyp gesetzt
  - active_at wird auf das period_invoice_template.start_on gesetzt
- Der GroupsFilter braucht keine Relation zur Sammelrechnung

Detail-Begründungen:
- Eine Einfachauswahl ist nach aktuellem Stand (für die deutschen Pfadis) genug, solange in der Gruppenhierarchie tief gesucht wird (z.B. Stämme können sowohl direkt an Landesverbänden als auch an Bezirken in Landesverbänden hängen). Mit dem GroupsFilter Konstrukt können wir das in Zukunft auch noch erweitern.
- Dass wir dies via Datenbankrelation zu GroupsFilter umsetzen statt nur einen `recipient_group_type` zu speichern und den Filter on the fly zu konstruieren, ist Entscheidung von der Besprechung am Daily am 9.1.26

# ToDo

- [x] Auswahl-Select (Dropdown-Menü) für `recipient_group_type` implementieren, welches die verfügbaren Gruppentypen gemäss Tech Spec zur Auswahl anbietet
- [x] Erklärenden hartcodierten Hilfe-Text unter dem Select einfügen: "Alle Gruppen von diesem Typ in oder unterhalb `#{layer.name}` bekommen eine Rechnung. Ausgenommen sind Gruppen, welche bereits beim Startdatum gelöscht oder archiviert waren."
- [x] Bei Änderung der Auswahl müssen die Felder der PeriodInvoiceTemplate::Items via XHR neu geladen werden, da sich z.B. die verfügbaren Rollentypen ändern können
- [x] Beim Speichern von PeriodInvoiceTemplates den recipient_groups_filter anlegen oder aktualisieren, mit Attributen gemäss Tech Spec
- [x] Sobald mindestens ein Rechnungslauf in der Sammelrechnung existiert, kann der Empfänger-Gruppentyp nicht mehr geändert werden
- [x] Specs schreiben
- [x] Mit angemessener Rolle "durchklicken"
- [x] [DoD](https://github.com/hitobito/hitobito/blob/master/doc/developer/process/definition_of_done.md) geprüft und erfüllt?
- [ ] CHANGELOG-Eintrag unter "unreleased" unten hinzufügen

carlobeltrame:
Fixes #3749
Please review commit by commit. Commits are already squashed to the minimal amount.

### DIFF
--- a/app/views/period_invoice_templates/_form.html.haml
+++ b/app/views/period_invoice_templates/_form.html.haml
@@ -7,7 +7,7 @@
   = hidden_field_tag :autosubmit
   = f.labeled_input_fields :name, :start_on, :end_on
 
-  = f.labeled_collection_select(:recipient_group_type, entry.group.class.child_types, :to_s, ->(group_type) { group_type.model_name.human(count: 2) }, {}, class: "form-select form-select-sm", disabled: entry.invoice_runs.any?, help_inline: t(".recipient_group_type_help", group: entry.group), data: { action: "change->autosubmit#save", turbo_frame: :period_invoice_template_items })
+  = f.labeled_collection_select(:recipient_group_type, entry.group.class.child_types, :to_s, ->(group_type) { group_type.model_name.human(count: 2) }, {}, class: "form-select form-select-sm", disabled: entry.invoice_runs.any?, help_inline: t(".recipient_group_type_help", group: entry.group), data: { action: "change->autosubmit#submit", turbo_frame: :period_invoice_template_items })
 
   = turbo_frame_tag :period_invoice_template_items, target: :_top do
     = f.labeled t("activerecord.attributes.period_invoice_template.items"), mark_as_required: true, label_class: "col-md-3 col-xl-2 align-self-baseline", class: "align-self-baseline" do
--- a/app/javascript/controllers/autosubmit_controller.js
+++ b/app/javascript/controllers/autosubmit_controller.js
@@ -15,24 +15,24 @@ export default class extends Controller {
   };
 
   initialize() {
-    this.save = this.save.bind(this);
+    this.submit = this.submit.bind(this);
   }
 
   connect() {
     if (this.delayValue > 0) {
-      this.save = debounce(this.save, this.delayValue);
+      this.submit = debounce(this.submit, this.delayValue);
     }
   }
 
   clear() {
-    const elem = this.element.querySelector("[data-action='autosubmit#save']");
+    const elem = this.element.querySelector("[data-action='autosubmit#submit']");
     if(elem) {
       elem.value = "";
       this.element.requestSubmit();
     }
   }
 
-  save(event) {
+  submit(event) {
     const submit = document.querySelector("input[name=autosubmit]");
     this.#withAutosubmitValue(submit, event.target.name || "autosubmit", () => {
       this.#withTurboFrame(event.target.dataset.turboFrame, () => {
@@ -61,7 +61,7 @@ export default class extends Controller {
    * is the case when the form is both used normally for submitting CRUD requests, but also
    * some inputs should trigger a partial reload of the form (because some other fields might
    * depend on the values in the previous fields). In that case, we can specify the
-   * data-turbo-frame value on the element triggering the autosubmit#save action.
+   * data-turbo-frame value on the element triggering the autosubmit#submit action.
    */
   #withTurboFrame(turboFrameId, callback) {
     if (!turboFrameId) return callback();
--- a/app/views/households/_edit_form.html.haml
+++ b/app/views/households/_edit_form.html.haml
@@ -10,6 +10,6 @@
     class: 'form-control form-control-sm person-search',
     placeholder: I18n.t('households.edit.placeholder_search_person'),
     autofocus: true,
-    data: { controller: 'tom-select', action: 'autosubmit#save',
+    data: { controller: 'tom-select', action: 'autosubmit#submit',
     tom_select_url_value: query_household_path(person_id: person.id),
     tom_select_no_results_value: I18n.t('global.no_list_entries')})
--- a/app/views/shared/_search_form.html.haml
+++ b/app/views/shared/_search_form.html.haml
@@ -7,7 +7,7 @@
   = hidden_field_tag :returning, true
   = hidden_field_tag :page, 1
   .has-clear.input-group
-    = search_field_tag :q, params[:q], class: 'form-control form-control-sm', placeholder: t('global.search.placeholder_person'), data: { action: "autosubmit#save"}
+    = search_field_tag :q, params[:q], class: 'form-control form-control-sm', placeholder: t('global.search.placeholder_person'), data: { action: "autosubmit#submit"}
     %span.input-group-text.fw-bold
       = icon(:spinner, class: "fa-xs fa-spin d-none") # gets is displays via form[aria-busy] toggle
       = icon(:times, class: "fa-xs", role: :button, data: { action: "click->autosubmit#clear" })
--- a/spec/views/households/edit.html.haml_spec.rb
+++ b/spec/views/households/edit.html.haml_spec.rb
@@ -110,7 +110,7 @@ describe "households/edit.html.haml" do
 
     it "has tom select field for adding new member" do
       expect(tom_select["autofocus"]).to eq "autofocus"
-      expect(tom_select["data-action"]).to eq "autosubmit#save"
+      expect(tom_select["data-action"]).to eq "autosubmit#submit"
       expect(tom_select["data-controller"]).to eq "tom-select"
       expect(tom_select["data-tom-select-url-value"]).to eq query_household_path(person_id: 2)
       expect(tom_select["data-tom-select-no-results-value"]).to eq "Keine Einträge gefunden."


app/controllers/period_invoice_templates_controller.rb
```diff
@@ -6,12 +6,20 @@
...
       items_attributes: [
         :id, :type, :name, :cost_center, :account, :vat_rate, :_destroy,
         dynamic_cost_parameters: [:unit_cost] +
           Settings.groups.period_invoice_templates.item_classes.to_h.values.flatten
       ]
     }]
+
+  def assign_attributes
+    super
+    entry.recipient_source ||= GroupsFilter.new
```
njaeggi, line 38: Wollen wir dafür nicht den `InvoiceRuns::RecipientSourceBuilder` erweitern und verwenden wie wir es bereits im `InvoiceRunsController` machen?

app/helpers/period_invoice_templates_helper.rb
```diff
@@ -0,0 +1,45 @@
+# frozen_string_literal: true
+
+#  Copyright (c) 2022, Pfadibewegung Schweiz. This file is part of
```
njaeggi:
```suggestion
#  Copyright (c) 2026, Hitobito AG. This file is part of
```

app/helpers/period_invoice_templates_helper.rb
```diff
@@ -0,0 +1,45 @@
...
+#  hitobito and licensed under the Affero General Public License version 3
+#  or later. See the COPYING file at the top-level directory or at
+#  https://github.com/hitobito/hitobito.
+
+module PeriodInvoiceTemplatesHelper
+  def format_recipient_group_type(period_invoice_template)
+    period_invoice_template.recipient_group_type.constantize.model_name.human(count: 2)
+  end
+
+  def format_calendar_tags(calendar, excluded:)
```
njaeggi: This is exactly the same as `format_calendar_tags` from `CalendarsHelper`. can't we use that instead?

app/helpers/period_invoice_templates_helper.rb
```diff
@@ -0,0 +1,45 @@
...
+      "calendars.tags.explanation_#{exclusion}",
+      count: tags.count,
+      tags: format_tag_list(tags),
+      events: format_event_type_all(calendar.group)
+    )
+  end
+
+  private
+
+  def format_event_type(group, event_type)
```
njaeggi: Same with this, also for the next two methods.... was this forgotten when copying the file or something?

app/models/period_invoice_template.rb
```diff
@@ -18,8 +19,26 @@ class PeriodInvoiceTemplate < ActiveRecord::Base
     on_or_after: :start_on,
     on_or_after_message: :must_be_later_than_start_on,
     if: -> { start_on.present? }
+  validates :recipient_group_type, presence: true, inclusion: {in: ->(entry) {
+    entry.group.class.child_types.map(&:name)
+  }}
+  validate :assert_changes_to_recipient_group_allowed
+
+  after_save :save_recipient_source
```
njaeggi: I think we could use `autosave: true` on our belongs_to instead of saving in after_save

app/views/period_invoice_templates/_form.html.haml
```diff
@@ -3,18 +3,20 @@
 -#  or later. See the COPYING file at the top-level directory or at
 -#  https://github.com/hitobito/hitobito.
 
-= entry_form(buttons_top: false, data: { controller: "period-invoice-template-form", period_invoice_template_form_wrapper_selector_value: "#items_fields .fields" }) do |f|
+= entry_form(buttons_top: false, data: { controller: "period-invoice-template-form autosubmit", period_invoice_template_form_wrapper_selector_value: "#items_fields .fields" }) do |f|
```
njaeggi, line 6: I don't really know if the autosubmit controller is correct here, it does what it should, but is a submit (POST) really needed here, We only get new options for the role count item dropdown... I think a GET request that would actually return 200 instead of 422 would be more suitable here, then we also wouldn't have to override create and update specifically for this.

This implementation isn't "wrong" in my head GET requests and a new stimulus controller would just make more sense 🙃

njaeggi (changes requested):

app/controllers/period_invoice_templates_controller.rb
carlobeltrame: Habe ich auch überlegt, aber ich habe hier keine passenden Params, bzw. sind die Params sehr period_invoice_template spezifisch, und ich müsste zerbrechliche Heuristiken anwenden, um aus den Params herauszulesen dass wir hier im PeriodInvoiceController sind und darum etwas ganz anderes machen müssen als sonst. Es scheint mir falsch, dass diese generische Klasse RecipientSourceBuilder die Parameter-Formate diverser Controller kennen muss.
Ich finde den RecipientSourceBuilder im vielseitig verwendbaren InvoiceRunsController nützlich um zu parsen, welche Personenliste, Teilnahmeliste oder Mailingliste die recipient_source ist. Aber in diesem neuen PeriodInvoiceTemplatesController gibt es nur eine Quelle für die recipient_source, nämlich die Einstellungen aus dem Formular selber, und es braucht keinen Mechanismus der Werte aus den Query Parameters parst (solange wir nicht z.B. erlauben, auf einer Personenliste oder Mailingliste direkt eine Sammelrechnung zu definieren).

app/helpers/period_invoice_templates_helper.rb
carlobeltrame: Yes, all of this can be removed. Thanks!

app/models/period_invoice_template.rb
carlobeltrame: Good idea, I'll try that :+1:

app/views/period_invoice_templates/_form.html.haml
carlobeltrame: There are a few issues here:
- the POST request
- the status 422
- reuse of the `create`/`update` actions for just fetching a refreshed form state (which normally the `new`/`edit` actions are responsible for)
- the autosubmit controller.

I'll address them separately.

The problem is that this is a form which is supposed to be heavily customized in wagons (with custom invoice items), and when using a GET request all the params of the whole form are appended to the URL, which can run into a limit which we would not easily notice. So I feel like a POST request is better.

As for status 200 instead of 422, I could try harder to implement that (I did try already but couldn't get it to work). The problem is that on the `edit` form, dry_crud does not assign_attributes, whereas on the `new` form it does. But I could probably add this and have the HTTP status be 200.

If we want a POST request but a status of 200, do you think we should use separate routes for this request? I.e. would it be better to separate this "give me a refreshed form state with all my inputs" into a separate controller action? If so, how would we name this new action? Having a separate action would make it harder to reuse the autosubmit stimulus controller.

I too am not completely convinced that the autosubmit stimulus controller is the right choice here, but it is already used (and was written for) partial submits of complex, multi-step forms in the steps_component, and then in some GET forms with limited params (household edit and people list search) to submit data which is also not persisted to the database. So in my opinion, the `#save` action of the autosubmit stimulus controller is misnamed, and if we renamed it to `#submit`, it would become clearer what it's used for.

I was actually already writing a separate stimulus controller for this form, when I noticed that the logic is very similar to the existing autosubmit controller, and since that one was already used so liberally, I thought DRY might be preferable.

What do you think?

njaeggi: The first issue about many custom params is also the thing I thought about, it makes sense to use a POST request here then. I would prefer the https status to be 200 tho, also on a post request, because updating the dropdown was succesful, not unprocessable or something was invalid. I don't know exactly what we have to change to make the response be 200, probably assign_attributes you mentioned seems like a good idea.

I also like the idea of a seperate route for this, be it the same controller and a custom method name, maybe something like period_invoice_templates_controller#invoice_items or a new controller period_invoice_templates/invoice_items_controller#create with a create method. Even tho invoice_items_controller isn't really the best name because we also have the invoice_item model... I can't really think of a better name right now tho.

About the stimulus controller, you are right, we use it the same way already in other places, #save would be a better name for the method tho, you are right. I was just looking at the controller with the GET request in my mind and was like yeah we don't submit anything here so we shouldn't use it here, but imo with a POST request and renaming the method to save, it would be way better already :rocket:

carlobeltrame: So this is what I did now:
- the POST request - still a POST request, in order not to put too many parameters in a GET request
- the status 422 - changed to status 200
- reuse of the `create`/`update` actions for just fetching a refreshed form state - still the same, because otherwise the autosubmit controller would also need to be extended to submit to these different routes only during autosubmit. If we do change it, I'd propose action names like `refresh_new` or `reload_edit` etc. @njaeggi hit me up if you think that would make the code a lot more readable. If we want to go that way, I'm not sure the autosubmit controller is still the right thing to use.
- the autosubmit controller - renamed the `#save` method to `#submit`, also in hitobito/hitobito_sac_cas#2341

PR #3998, opened 2026-03-03
Use UETR number to identify unique payments

njaeggi:
We can use UETR (Unique End-to-End Transaction Reference) to identify payments, UETR is a unique reference generated at the start of a payment and is not changed accross all providers. It is mandatory in all SWIFT payment networks since november 2020. Since ISO 20022 it is now a core standard for almost all modern payment reporting. In case of the UETR not exsting, we fallback to our old transaction identifier


spec/fixtures/invoices/camt.054_version_001.04.xml
njaeggi, line 1: This is the old standard, without the UETR number (same fixture file as before)

spec/fixtures/invoices/camt.054_version_001.08.xml
njaeggi, line 1: File with the same mock data in the new standard

PR #4023, opened 2026-03-13
Fix locale-preserving redirect after sign-in

Visualstudiocodetest:
When signing in from a locale-specific login URL (for example /fr/users/sign_in), Devise may reuse a stale stored return path in another locale. This change normalizes redirect locale in both path and query so the post-login redirect matches the requested sign-in locale.

### DIFF
--- a/app/controllers/devise/hitobito/sessions_controller.rb
+++ b/app/controllers/devise/hitobito/sessions_controller.rb
@@ -54,7 +54,7 @@ class Devise::Hitobito::SessionsController < Devise::SessionsController
     locale = requested_locale
     return path if locale.blank?
 
-    # Normalize both path and query locale so all redirect styles are covered.
+    # Normalize both path and query locale, because our root_path passes the locale as query param
     uri = URI.parse(path)
     uri.path = localize_path(uri.path, locale)
     uri.query = localize_query(uri.query, locale)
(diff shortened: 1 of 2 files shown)


app/controllers/devise/hitobito/sessions_controller.rb
```diff
@@ -42,6 +42,53 @@ def configure_permitted_parameters
...
+  def localize_stored_redirect(path)
+    return path if path.blank?
+
+    locale = requested_locale
+    return path if locale.blank?
+
+    # Normalize both path and query locale so all redirect styles are covered.
+    uri = URI.parse(path)
+    uri.path = localize_path(uri.path, locale)
+    uri.query = localize_query(uri.query, locale)
```
carlobeltrame, line 60: Hitobito does not use a locale query param. The localize_query logic can be deleted.

And then, we don't need to parse the URI anymore. The string manipulation localize_path can directly be applied to `path`. You can even do the `.sub` call directly here, without going through the extra `localize_path` method with another early exit return.

app/controllers/devise/hitobito/sessions_controller.rb
```diff
@@ -42,6 +42,53 @@ def configure_permitted_parameters
...
+  end
+
+  def requested_locale
+    locale = params[:locale].to_s
+    locale if configured_locales.include?(locale)
+  end
+
+  def configured_locales
+    # Keep this dynamic so wagon/core language settings are respected.
+    @configured_locales ||= Settings.application.languages.to_hash.keys.map(&:to_s)
```
carlobeltrame, line 72: The `to_hash` call is not needed. The rest is fine.

carlobeltrame:
A feature spec (end-to-end test) for this would be nice, so it doesn't get broken in the future.

PR #4024, opened 2026-03-13
Enforce 4-digit year on date inputs

Visualstudiocodetest:
Summary
- enforce a 4-digit year pattern on shared date input helpers
- apply the same placeholder/pattern defaults to filter date inputs
- update role start/end date fields to use explicit 4-digit year guidance

## Notes
- this keeps existing date format behavior and only tightens the year to 4 digits"


Visualstudiocodetest (commented):
Making the check in the server side and added translation for the error message

app/models/role.rb
```diff
@@ -314,6 +316,16 @@ def assert_type_is_allowed_for_group
     end
   end
 
+  def date_year_must_be_four_digits
+    [:start_on, :end_on].each do |attribute|
+      date = public_send(attribute)
+      next if date.blank?
+      next if date.year.abs <= 9999
```
Visualstudiocodetest: add minimum 1900 year

Visualstudiocodetest: Do the dates for camps and, courses ,events ,date of birth

Visualstudiocodetest (commented):
Fixed schema.rb whith the correct date time so all the migrations are runned

PR #4028, opened 2026-03-14
Show pending applications only once in profile

fhinok:
Change query to only include pending applications.
Upcoming participations are already queried in upcoming_participations.

### DIFF
--- a/spec/decorators/person_decorator_spec.rb
+++ b/spec/decorators/person_decorator_spec.rb
@@ -100,7 +100,10 @@ describe PersonDecorator, :draper_with_helpers do
 
   context "participations" do
     it "pending_applications returns participations that are not active" do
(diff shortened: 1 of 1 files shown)


app/domain/person/event_queries.rb
```diff
@@ -13,9 +13,7 @@ def initialize(person)
   def pending_applications
     person.event_applications
       .merge(Event::Participation.pending)
-      .merge(Event::Participation.upcoming)
```
carlobeltrame, line 16: You are removing the filter for upcoming applications. This means that newly, past applications will also show up. I think that is not what we want here... I would have suggested a distinct, but that is already applied to this query. Probably the DISTINCT does not filter the duplicates because of the select event_dates.start_at ? Can you try removing event_dates.start_at instead of this change?

PR #4034, opened 2026-03-14
Fix IMAP spam tab crash when mailbox contains thousands of messages

patrickuhlmann:
Previously, `Imap::Connector#fetch_mails` loaded all messages from the selected mailbox at once using a sequence-number range fetch, including the full RFC822 body of every message. With 13,000+ spam messages this caused the application to crash due to exhausting available memory, before Kaminari could even paginate the result.

Fix this by replacing the single bulk fetch with a two-phase approach within one IMAP connection:
1. `uid_search(["ALL"])` retrieves only message UIDs — a lightweight list of integers regardless of mailbox size.
2. UIDs are sorted descending (newest first) and sliced to the current page.
3. `uid_fetch` retrieves the full message content for only the ~25 messages on the current page.

The old bulk-fetch logic is retained as `fetch_all_mails`, used by `AppStatus::Mail` which must inspect every inbox message to detect overdue (stuck) deliveries.

Fixes: https://github.com/hitobito/hitobito/issues/3549

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Version 2.8
 
 - Mail Inboxes Ansicht funktioniert jetzt auch bei einer grösseren Anzahl Emails (#3549 merci @patrickuhlmann)
+- Abgrenzung aktiver von gelöschten Gruppen in der "Rollen die auf mich Zugriff haben" Ansicht (hitobito/hitobito_cevi#200, merci @patrickuhlmann)
 - Labels von Kontaktinformationen (Telefonnummern etc.) können auf die vordefinierten Werde eingeschränkt werden. In den Personenexporten werden pro vordefiniertem Label eine Spalte ausgegeben (#1800) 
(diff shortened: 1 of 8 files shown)


app/utils/imap/connector.rb
```diff
@@ -68,6 +81,14 @@ def config(key)
 
   private
 
+  def fetch_mails_by_uids(page_uids)
+    return [] if page_uids.empty?
+
+    fetch_data = @imap.uid_fetch(page_uids, attributes)
+    mails = fetch_data.blank? ? [] : fetch_data.map { |mail| Imap::Mail.build(mail) }
+    mails.sort! { |a, b| b.date.to_i <=> a.date.to_i }
```
amaierhofer: Das sort hat bei meinem tests mit er pbs prod nicht immer  funktioniert, da der der `envelope` der Mail mitunter `nil` war. Ist wahrscheinlich ein edge case aber müssen wir abfangen. Am einfachsten direkt in der `Mail`

Und dann fände ich noch schön, wenn die `Mail` selbst sortierbar wäre, dort also den comparison operator `<=>` implementieren.

patrickuhlmann: Hast du schon versucht zu pushen. Hatte schon von Anfang an "Allow edits and access to secrets by maintainers" gesetzt auf dem PR:
image

amaierhofer: Stimmt. das Problem lag bei mir, push hat geklappt, danke!

PR #4044, opened 2026-03-15
Fix: additional emails invoice mailings flag api

B52-Bomber:
The invoice and mailings flags of additional_emails are now included in API responses.

Issue: https://github.com/hitobito/hitobito_pbs/issues/348


ThomasEllenberger:
Because of the upcoming release, we postpone the merging of this pull request until after the release.
API changes are something we like to inform our customers as much in advance as possible.

PR #4055, opened 2026-03-18
Allow to update attributes of CustomContents with seeds

kronn:
We switched to seed_once back when we used this to seed the intial
content. This prevented the user-changes being overwritten by
a deployment, while still allowing to seed new CustomContents. We now
use and suggest to run 'rake db:seed wagon:seed' on every deployment.

After we introduced globalize to support translations by the users,
stored in the database, we split the seeding into two parts: first the
attributes like key and placeholders, then the translations. The
seed-once is sensible, even essentials to not destroy user-input of the
translations. For the attributes is does not make sense anymore.

Therefore, I switch this to normal seeding, allowing us to change the
placeholders being used by a CustomContent if needed.

### DIFF
--- a/spec/mailers/hitobito_log_mailer_spec.rb
+++ b/spec/mailers/hitobito_log_mailer_spec.rb
@@ -20,11 +20,27 @@ describe HitobitoLogMailer do
   end
 
   it "renders for log entry for mailing list" do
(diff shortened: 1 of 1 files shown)


daniel-illi (approved):
LGTM 👍

PR #4082, opened 2026-03-24
Feature/3503 Improve toggle button screen reader focus styling

### LINKED ISSUE #3503: Tastaturbedienung - Fokus für Toggle-Buttons
Die Toggle-BUttons (Switches) haben heute keinen Fokus-Style. Dieser soll eingebaut werden, damit man bei der Bedienung per Tastatur erkennt, wo der Fokus ist.

Vgl. Bild:

Image

RandomTannenbaum:


app/javascript/stylesheets/hitobito/_form.scss
```diff
@@ -603,8 +609,9 @@ input.switcher {
     text-indent: -5000px;
     height: 20px;
     width: 40px;
-    top: 10px;
```
RandomTannenbaum, line 606: This was probably set to align the switchers with the markers of the unordered list on the person add requests page (see next comment). At the same time this lead to a misalignment of the switcher on the label formats settings page (the only other page where the switcher is in use currently). I now use flexbox to align the switchers.

app/javascript/stylesheets/hitobito/_layout.scss
```diff
@@ -417,8 +417,14 @@ strong.bigger-title div.trix-content {
   margin-bottom: 5px;
 }
 
-.approvers-list label {
-  margin-bottom: 4px;
+.approvers-list li::before {
```
RandomTannenbaum: On the `group/person_add_requests/_approvers` partial, there is an unordered list with each list item containing a switcher. I assume that the earlier mentioned `top: 10px` was used, to align the switchers with the markers of the list items. I removed this and am now using Flexbox to vertically align the content of the list items. The problem is, that `<li>` elements have the property `display: list-item`. Setting `display: flex` overrides this, resulting in the list items having no marker.

So far I haven't found another suitable solution for aligning the marker with list item content except using the `::before` pseudo-class to create a custom marker as for example also suggested [here](https://stackoverflow.com/questions/69874236/how-can-i-vertically-align-a-list-item-marker). There they use an empty content, set fixed width and height and round it with `border-radius: 50%`. This works but for whatever reason, the circles are not always perfectly round and sometimes have some pixels missing which makes them look wonky and uneven. I couldn't figure out why this is so I resorted to using the Unicode bullet point, which also seems to be the standard character used in unordered lists. This has the advantage of being consistent and round but has the drawback of not having the same size in all browsers, because Unicode characters have different sizes with different fonts.

There also exists the [`::marker` pseudo class](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Selectors/::marker), however it is not widely supported.

The current solution works and is aligned but seems a bit hacky to me. On the other hand most solutions I've seen online have suggested a similar approach. Maybe we should just completely remove the unordered list or just leave it as is for now.

carlobeltrame: I don't see any reason why we need the bullet points. I simply removed them.

carlobeltrame (approved):
Habs im UI getestet, aber habe keinen Screenreader bei mir eingerichtet. Sieht so weit gut aus für mich :+1:

PR #4103, opened 2026-03-30
Ruby 4.0

### LINKED ISSUE #3689: BUILD: Ruby 3.4
# Ausgangslage
Aktuell verwenden wir Ruby 3.2

# Anforderungen
Ruby 3.2 hat EOL 2026-03-31, daher wollen wir auf Ruby 3.4 aktualisieren.

# Tech-Spec

- Ruby aktualisieren
- Gem-upgrades soweit nötig

# ToDo

- [ ] Dockerfile anpassen
- [ ] bin/active_wagon.rb anpassen
- [ ] development-setup anpassen
- [ ] Integrationen testen
- [ ] alle Wagons durchtesten
- [ ] [DoD](https://github.com/hitobito/hitobito/blob/master/doc/developer/process/definition_of_done.md) geprüft und erfüllt?
- [ ] CHANGELOG-Eintrag unter "unreleased" unten hinzufügen

kronn:
This needs a second look after being run on CI. I ran into some persistant problems locally and want to rule out a broken setup first. Also, CI itself needs to run with the new version...


rnestler:
I'd suggest to transition to the latest 3.2, then to 3.3 and then stay with 3.4 which is the recommended Ruby version for Rails 7.2, 8.0 and 8.1. See https://www.fastruby.io/blog/ruby/rails/versions/compatibility-table.html

kronn:
We always update patch-levels once they are released. Given the date of the publication, Ruby 3.5 / 4.0 are not even in the scope of that comparison. For their business model, this makes sense.

Since we have a lot of specs covering many aspects of the application, I would go for the newer version. The changelogs for the newer version do not mention breaking changes that were not resolved earlier anyway. (See: net-http 0.7)

PR #4154, opened 2026-04-21
Allow to use Redis for caching

rnestler:
Open questions:
 * [x] Can/should I just add the redis gem unconditionally to the Gemfile? Alternatively I could add it conditionally behind an env var guard or just leave it up to the wagon to add the gem. `redis` gem got added in #4092.
 * [ ] Example configuration

### DIFF
--- a/config/application.rb
+++ b/config/application.rb
@@ -111,8 +111,13 @@ module Hitobito
 
     config.middleware.insert_before Rack::ETag, Rack::Deflater
 
-    config.cache_store = :mem_cache_store, { compress: true,
+    if ENV["RAILS_CACHE_REDIS_URL"].present?
+      additional_redis_options = YAML.load("{ #{ENV['RAILS_CACHE_REDIS_CONFIG']} }").symbolize_keys
+      config.cache_store = :redis_cache_store, { url: ENV["RAILS_CACHE_REDIS_URL"] }.merge(additional_redis_options)
+    else
+      config.cache_store = :mem_cache_store, { compress: true,
                                              namespace: ENV['RAILS_HOST_NAME'] || 'hitobito' }
+    end
 
     config.active_storage.variant_processor = :vips
 
--- a/doc/operator/01_deployment.md
+++ b/doc/operator/01_deployment.md
@@ -34,14 +34,14 @@ gesetzt werden. Werte ohne Default müssen in der Regel definiert werden.
 | RAILS_MAIL_DOMAIN | Der Domainname für die Mailinglisten/Abos | `RAILS_HOST_NAME`        |
 | RAILS_MAIL_RETRIEVER_TYPE | `pop3` oder `imap`, alles was vom [Mail](https://github.com/mikel/mail) Gem unterstützt wird. | `pop3`                   |
 | RAILS_MAIL_RETRIEVER_CONFIG | Eine Komma-separierte `key: value` Liste mit allen erforderlichen E-Mail Empfangseinstellungen des gewählten Typs, z.B. `address: mailhost.local, port: 995, enable_ssl: true`. Siehe [Mail](https://github.com/mikel/mail#getting-emails-from-a-pop-server) für gültige Optionen. Wenn diese Variable nicht gesetzt ist, funktionieren die Mailinglisten nicht. | -                        |
-| RAILS_CACHE_REDIS_URL | Optional: URL zum redis kompatiblen server welcher für caching verwendet werden soll | - |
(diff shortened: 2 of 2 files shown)


kronn (commented):
In general, I like the idea of an easier switch to Redis. 

We have also this PR: https://github.com/hitobito/hitobito/pull/4092/ and I would like them to be compatible. @RandomTannenbaum is working on the other one. Could you coordinate that this can coexist? I will add him as reviewer for this purpose.

The pattern of having YAML in an ENV-Var is not ideal, especially if passwords, usernames or generally quotes are involved. I would like to see a documented example how this is expected to be configured.

rnestler:
> In general, I like the idea of an easier switch to Redis.

Glad to hear :slightly_smiling_face: 

> We have also this PR: https://github.com/hitobito/hitobito/pull/4092 and I would like them to be compatible. @RandomTannenbaum is working on the other one. Could you coordinate that this can coexist? I will add him as reviewer for this purpose.

Sure. With the `REDIS_URL` you can specify the server, credentials and the "database" number to connect to:
```
REDIS_URL=rediss://:{PASSWORD}@{FQDN}:{port}/{db_index}
REDIS_URL=rediss://foobar@my.redis.example:6379/0
```

So we could use "#{REDIS_URL}/0" for cache and "#{REDIS_URL}/1"  for cable.

> The pattern of having YAML in an ENV-Var is not ideal, especially if passwords, usernames or generally quotes are involved. I would like to see a documented example how this is expected to be configured.

Yeah I also didn't like it, but sometimes one will need to set additional connect options.

RandomTannenbaum:
> Sure. With the `REDIS_URL` you can specify the server, credentials and the "database" number to connect to:
> 
> ```
> REDIS_URL=rediss://:{PASSWORD}@{FQDN}:{port}/{db_index}
> REDIS_URL=rediss://foobar@my.redis.example:6379/0
> ```
> 
> So we could use "#{REDIS_URL}/0" for cache and "#{REDIS_URL}/1" for cable.

I haven't really looked into the specifics of the Redis configuration so far but as far as I can tell what @rnestler suggested should indeed be compatible with what I am building :slightly_smiling_face:.

The only caveat I can think of right now is, that it wouldn't be possible to use Memcached for caching and Redis for ActionCable at the same time if they both use the same ENV var for the URL. Idk if anyone would wanna do that though. @kronn Do you already have plans on how and when you want to switch over the Cache?

config/application.rb
```diff
@@ -111,8 +111,13 @@ class Application < Rails::Application
 
     config.middleware.insert_before Rack::ETag, Rack::Deflater
 
-    config.cache_store = :mem_cache_store, { compress: true,
+    if ENV["REDIS_URL"].present?
+      additonal_redis_options = YAML.load("{ #{ENV['RAILS_REDIS_CACHE_CONFIG']} }").symbolize_keys
```
RandomTannenbaum:
```suggestion
      additional_redis_options = YAML.load("{ #{ENV['RAILS_REDIS_CACHE_CONFIG']} }").symbolize_keys
```

Small typo 🙂

config/application.rb
```diff
@@ -111,8 +111,13 @@ class Application < Rails::Application
 
     config.middleware.insert_before Rack::ETag, Rack::Deflater
 
-    config.cache_store = :mem_cache_store, { compress: true,
+    if ENV["REDIS_URL"].present?
```
kronn: In order to have more fine-grained control, I suggest using `RAILS_CACHE_REDIS_URL` and `RAILS_CACHE_REDIS_CONFIG` as environement variables. The used DB can and should then be encoded in the provided URL, to allow for other valid setups.

kronn (changes requested):
I would rename the environment variables to avoid conflicts with other redis-uses.

config/application.rb
rnestler: Fixed.

config/application.rb
rnestler: Fixed.

kronn:
I totally forgot or overlooked to adress the first question: yes, I would add the redis-gem unconditionally. Firstly, we plan to use it anyway. Secondly, this would be in line of having multiple error-reporting possibilities. While we do not use it anymore, the airbrake-gem is still present, alongside sentry.

At some point we will drop those unused parts and document ways to keep them if needed. But that is a task for future-us.

But, the other mentioned PR will add the redis-gem already. So, not you do not need to do that. It will however happen :-)

For this PR, I would first address the PR for the "UI for background Jobs". After that is merged, I will rebase and merge this one.

kronn (approved):

PR #4156, opened 2026-04-21
Generate a filename, even if no invoices are exported

kronn:
It might happen for a variety of reasons, some of which are legit and the others the wish of a user. Either way, presenting a 500 is more expensive than returning an empty file, mostly due to support requests and error-tracking.

Fixes https://glitchtip.puzzle.ch/hitobito/issues/14460

**Maybe** this needs to be a hotfix, should be possible easily enough. the fix has been tested against master and stable.


daniel-illi (commented):
LGTM 👍

kronn:
merged into stable :heavy_check_mark:

PR #4160, opened 2026-04-23
Do not crash bin/wagon gemfile if BUNDLE_GEMFILE is not set

rnestler:
While it has an if clause to guard the process would exit anyways since `set -u` was set. By substituting a default we fix this.

### DIFF
--- a/bin/wagon
+++ b/bin/wagon
@@ -198,7 +198,7 @@ gemfile)
   bundle check || bundle install
   $0 configs Gemfile.lock
 
-  if [ -n "${BUNDLE_GEMFILE}" ]; then
+  if [ -n "${BUNDLE_GEMFILE:-}" ]; then
     $0 configs "${BUNDLE_GEMFILE}.lock"
   fi
   ;;


bin/wagon
```diff
@@ -198,7 +198,7 @@ gemfile)
   bundle check || bundle install
   $0 configs Gemfile.lock
 
-  if [ -n "${BUNDLE_GEMFILE}" ]; then
+  if [ -n "${BUNDLE_GEMFILE:-}" ]; then
     $0 configs "${BUNDLE_GEMFILE}.lock"
```
rnestler, line 202: I'm confused a bit why this should be needed anyways: We don't want Gemfile.local to end up in the wagon, do we? @kronn Or at least it should be `.gitignore`d as well.

rnestler: At least the sac wagon ignores it https://github.com/hitobito/hitobito_sac_cas/blob/master/.gitignore

kronn: I have it in my global gitignore or so. This special BUNLDE_GEMFILE is one of these idiosyncracies I want to get rid of in the long run. it is used to avoid having an ever-changing Gemfile.lock when switching wagons.

kronn (approved):

PR #4163, opened 2026-04-23
Allow wagons not to be git submodules

rnestler:
Just `rm -r` would fail if the wagon isn't a git submodule.


Dockerfile
rnestler, line 31: I may still need to fix the content of `WAGON_VERSIONS` but I'm not sure what it is used for exactly.

kronn: The file is used to show information about the deployed image in the footer. if is is missing, nothing should be shown.

See https://github.com/hitobito/hitobito/blob/9e213fe2bb593771dc163fec4d60754798accb94/config/application.rb#L139-L143

kronn (approved):

PR #4175, opened 2026-04-28
Support reminding and sending emails for invoices only issued (not sent)

amaierhofer:
https://help.puzzle.ch/#ticket/zoom/11148


daniel-illi (approved):
LGTM 👍

PR #4210, opened 2026-05-08
Improve Performance for deleted people queries

codez:


kronn (approved):
At first, I had worries that we are over-indexing the table. This is seemingly not the case as we cover very different queries.

PR #4215, opened 2026-05-08
Bug/Fix setting capybara driver inconsistently

RandomTannenbaum:
Capybara.current_driver is intended to be used to change the driver for
specific specs e.g. in a before hook. Because of this, the driver set
with current_driver is automatically reset to the standard driver after
a test runs runs. In our case this is the Rack::Test driver since we dont
specify Capybara.default_driver. Because we specified current_driver in the
spec_helper, which is loaded before all specs run, the first feature
spec would always run with the chrome driver, regardless of the presence
or rather absence of `js: true`. This would lead to seemingly flaky
specs when either the test only worked with the js driver and `js: true`
wasnt specified or the test for some reason only worked with the
Rack::Test driver.

Going forward, only test that explicitly define `js: true` should
actually run with the chrome driver.

### DIFF
--- a/app/models/paper_trail/version.rb
+++ b/app/models/paper_trail/version.rb
@@ -11,6 +11,7 @@
 #
 #  id             :integer          not null, primary key
 #  event          :string           not null
+#  item_label     :string
 #  item_subtype   :string
 #  item_type      :string           not null
 #  main_type      :string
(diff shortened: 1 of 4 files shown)


njaeggi (approved):
LGTM

PR #4229, opened 2026-05-12
Use .lft instead of .name for layer sorting

### LINKED ISSUE #1720: GROUPS: Gruppensortierung in Hierarchie
### Aktuelles Verhalten

Die Sortierung der Gruppen wird in einem `after_save` hook gemacht. Problem ist, dass schon bei mehreren Umgebungen eine Diskrepanz festgestellt wurde. 

### Gewünschtes Verhalten

Die Gruppen sollten immer korrekt einsortiert sein

### Schritte zu reproduzieren

Leider unklar wie der Fehler entsteht

_No response_

### hitobito version

1.27.13

### Organisation

Die Mitte

Nestedset Gem: https://github.com/collectiveidea/awesome_nested_set

### ToDos

Versuchen herauszufinden, Fehler zu reproduzieren:
- after_save Hook debuggen
- mit Timing spielen

rnestler:
This uses the sorting defined by sorting_name (which defaults to display_name).

I think this fixes #1720. Can @carlobeltrame or @nilsrauch confirm?

Before:
image

After:
image

### DIFF
--- a/app/models/group.rb
+++ b/app/models/group.rb
@@ -226,11 +226,11 @@ class Group < ActiveRecord::Base # rubocop:disable Metrics/ClassLength
     end
 
     # order groups based on order in Group.all_types
-    # group.name as second order attribute, to get same output for all
+    # group.lft as second order attribute, to get same output for all
     # queries where multiple groups have the same type
     def order_by_type
       joins("INNER JOIN group_type_orders ON group_type_orders.name = groups.type")
-        .reorder("group_type_orders.order_weight ASC, groups.name ASC")
+        .reorder("group_type_orders.order_weight ASC, groups.lft ASC")
     end
 
     private
--- a/app/views/groups/_child_groups.html.haml
+++ b/app/views/groups/_child_groups.html.haml
@@ -4,7 +4,7 @@
 -#  https://github.com/hitobito/hitobito.
 
 - @sub_groups.each do |label, groups|
-  = section_table(label, groups.sort_by(&:sorting_name)) do |item|
+  = section_table(label, groups) do |item|
     %td
       = link_to(item.to_s, group_path(item.id))
 


app/models/group.rb
```diff
@@ -226,11 +226,11 @@ def attr_used?(attr)
     end
 
     # order groups based on order in Group.all_types
-    # group.name as second order attribute, to get same output for all
+    # group.lft as second order attribute, to get same output for all
     # queries where multiple groups have the same type
     def order_by_type
       joins("INNER JOIN group_type_orders ON group_type_orders.name = groups.type")
-        .reorder("group_type_orders.order_weight ASC, groups.name ASC")
+        .reorder("group_type_orders.order_weight ASC, groups.lft ASC")
```
rnestler, line 233: Here it used only `name`, even if the actual sorting may be depending on additional fields as defined in `def sorting_name` of `app/models/group/nested_set.rb`

njaeggi (approved):

app/models/group.rb
njaeggi: Since the `short_name` is displayed and the after_save hook moves the group's `lft` and `rgt` based on `display_name` (short_name if present, otherwise name) this is the desired behaviour.

This also won't change something about the groups with `static_name = true` because these groups just return the class label instead of super in `name` and `short_name`

Nice catch :rocket:

rnestler:
@njaeggi are the wagon test failures flaky or related to the change?

njaeggi:
@rnestler The SAC Wagon Tests were broken last week due to development, I am trying to rerun the specs on this PR for once with the fixes in the SAC Wagon, the failures most likely aren't related to the fix.

And the reruns on wednesday last week did not really say a lot because Github Actions decided to stop on prepare steps for some reason for hours :upside_down_face:

njaeggi:
@rnestler Es waren noch kleine Anpassungen in zwei Wagons nötig, jedoch nicht Dinge die kaputt gehen sondern mehr Specs welche sich neu an das korrekte Sorting halten müsse :upside_down_face: 
- https://github.com/hitobito/hitobito_pbs/pull/470
- https://github.com/hitobito/hitobito_insieme/pull/215

PR #4310, opened 2026-06-16
Only paginate HTML responses

kronn:
Specifically, not exports


njaeggi (approved):
LGTM if specs don't break 🙃

nilsrauch (approved):

PR #4368, opened 2026-07-09
Preselect the correct label in translated dropdowns

kronn:
The internal value is the german locale, so we need to use that for comparing. ActionViews built-in FormBuilder does not know about this speciality, so we need to use the provided API to help it along.

One example: https://help.puzzle.ch/#ticket/zoom/11497


amaierhofer (approved):

kronn:

PR #4369, opened 2026-07-10
Some bounces may not look like it, but still have a code

kronn:
The code is in a MIME-Part of Type 'message/delivery-status'. If one such diagnostic code is found, it is a strong indicator for a bounce. The risk of false positive is lower than allowing bounces to be relayed


carlobeltrame (approved):

kronn:

PR #4396, opened 2026-07-26
Diagnostic-Codes are not recognized in reports

kronn:
Bounces are multi-part mails, these days. Since most mails are messages, the mail-gem only looks for the "Delivery Status Notification" in multipart-messages of type =~ /message/. In those messages, it looks for a part of the sub_type =~ /delivery-status/ and goes from there. This feels correct and handles the majority of cases.

However, there are multipart-messages with the main-type "report" that are (therefore) not technically bounces, but could and should be handled like them. Especially if they contain a part of type delivery-status.

So, we search the mails ourselves for status-notification-parts and try to determine bounces that way. In order to wrap this and keep the change localized, I removed the plain delegation of the diagnostic code to the mail-gem wrapped mail. Now, our Imap::Mail tries harder to determine the bounciness of a mail by looking for more diagnostic codes. It keeps the idiosyncracy of the mail-gem to return an array if there a multiple entries and the single entry if there only is one.

This extended detection of diagnostic codes might block handling mails if someone forwards a bounce to a mailing-list. Also, if the sender handcrafts a mail, uses the wrong main content-type and includes at least a valid subset of a delivery-status with a Diagnostic-Code, it will be handled like a bounce.
If this is a surprise, it should be a pleasent one.

### DIFF
--- a/app/utils/imap/mail.rb
+++ b/app/utils/imap/mail.rb
@@ -14,7 +14,7 @@ class Imap::Mail
   attr_accessor :net_imap_mail
 
   delegate :subject, :sender, to: :envelope
-  delegate :message_id, :final_recipient, :diagnostic_code, to: :mail
+  delegate :message_id, :final_recipient, to: :mail
 
   def self.build(net_imap_mail)
     entry = new
@@ -99,10 +99,28 @@ class Imap::Mail
   end
 
   def bounce?
-    mail.bounced? || mail.diagnostic_code.present?
+    mail.bounced? || diagnostic_code.present?
+  end
+
+  def diagnostic_code
+    mail.diagnostic_code.presence || all_diagnostic_codes
   end
 
   private
+
+  def all_diagnostic_codes
+    codes = delivery_status_parts.flat_map do |part|
+      # very abbreviated version of the mail-gem's #diagnostic_code, from Mail::Part
+      parsed = Mail::Header.new(part.raw_source)
+      Array(parsed["diagnostic-code"]).map(&:value)
+    end.compact
+
+    codes.one? ? codes.first : codes
+  end
+
+  def delivery_status_parts
+    mail.parts.select { |part| part.mime_type == "message/delivery-status" }
+  end
 
   def envelope
     @net_imap_mail.attr["ENVELOPE"]
--- a/spec/fixtures/email/helpful-return.eml
+++ b/spec/fixtures/email/helpful-return.eml
@@ -1,0 +1,213 @@
+Return-Path: <MAILER-DAEMON>
+X-Original-To: leaders@hitobito.example.com
+Delivered-To: leaders@hitobito.example.com
+Received: from mx03.hitobito.example.com (mx03.hitobito.example.com [45.81.71.124])
+	by mailstore02.cloud.hitobito.example.com (Postfix) with ESMTPS id 8D22972E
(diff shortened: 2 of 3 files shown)


nchiapol:
"This extended detection of diagnostic codes might block handling mails if someone forwards a bounce to a mailing-list." 

Not sure this would be a pleasant surprise. The most likely case for this is, forwarding of a mail to a support list run via Hitobito. At least the main support list for hitobito_cevi is independent of the whole hitobito setup - so we should not be affected.

kronn:
Well, forwarding a bounce to all people on the mailing-list would in most cases result in more bounces as the inbox which responded with the bounce is part of that list. We have this problem and this PR is intended to not overload mailboxes (again) through this mechanism. I am very happy for you, if this has not affected you.

For the case that someone wants to share an interesting bounce-message with friends, hitobito already does (and with this PR more often) just handle this as a bounce (not forward it). If someone uses hitobito to exchange such technical mails (which are normally automated and not meant to be forwarded) on a regular, we can surely build a feature to support just that.
In most cases, I think that hitobito mailing-lists are used to exchange other types of messages, like information about an upcoming event or so (i.e. normal e-mails). :-)

On a more serious note: Normal messages are not affected by this. The "Delivery Status Notification" in a mail is not meant to be forwarded. If this is needed, it should also not go to a multiplexing adress, but a single recipient. Support addresses and technical contacts will handle those accordingly. If absolutely needed, I suggest compressing the e-mail to be forwarded and attach it as a zip-file or so. At that point, it is a normal attachment. 

I see the point for giving support via E-Mail, but handling normal messages has higher priority. The unregulated and undetected forwarding of bounce-messages to everyone on a list brings far greater operational risks.

But: thanks for reading this.

amaierhofer (approved):

kronn:

nchiapol:
> I see the point for giving support via E-Mail, but handling normal messages has higher priority.

I agree, I just wanted to point out that there might be legitimate situations where something that contains typical "bounce-Strings" is actually not a bounce message. So this is a hack that introduces new unexpected behaviour in edge cases.

I think it would be a good idea to at least document this properly. (Ideally together with the rest of the bounce-handling)

kronn:
On Mon, Jul 27, 2026 at 03:08:58PM -0700, Nicola Chiapolini wrote:
> nchiapol left a comment (hitobito/hitobito#4396)
> 
> > I see the point for giving support via E-Mail, but handling normal messages has higher priority.
> 
> I agree, I just wanted to point out that there might be legitimate situations where something that contains typical "bounce-Strings" is actually not a bounce message. So this is a hack that introduces new unexpected behaviour in edge cases.

We are not only matching by "could not be delivered". Since bounce-mails
are multi-part mails (as are mails which have HTML and plain-text
versions, or mails with attachments), we are looking for a delivery
status report. This type of "attachment" is normally not present in
mails. There are different legitmate report-types, which should now be
all caught an analysed.

The only downside is that forwarding a bounce via
a hitobito-mailing-list MIGHT result in the it being parsed as bounce,
since some forwarding methods put the forwarded mails into a separate
part that would allow for the status-report to be extracted by this
change.

This results in hitobito being not the ideal tool for exchanging
bounce-messages. You still can (put the bounce into a text-file, zip it
and attach it), but it 

> I think it would be a good idea to at least document this properly. (Ideally together with the rest of the bounce-handling)

We are in the process of cleaning and improving the bounce- and
mail-handling. Once we reach a more stable version, I will document it.

PR #4426, opened 2026-08-19
Remove length restriction on PaymentReminderConfig#text (hitobito/hitobito_sww#414)

daniel-illi:


daniel-illi:
replaced by #4427

PR #4437, opened 2026-08-23
Use gender neutral wording for german translation

juliusstoerrle:
Um die Anzahl der angepassten Übersetzungen im pfadi_de Waggon zu reduzieren wird vorgeschlagen diese neutralen sprachlichen Anpassungen (Geschlechtsneutrale Formulierungen) in den Core zu überführen.


carlobeltrame (approved):
Danke!

juliusstoerrle:
Tests gefixed. 

Zu spec/controllers/events_controller_spec.rb folgende Erklärung: the checkbox view renders label_plural, but the test was asserting against .label (singular). That only worked by coincidence before because Helfer/-in was a substring of the old plural Helfer/-innen; the new plural Helfende broke that substring match. Fixed by asserting against .label_plural for all four roles, matching what's actually rendered.

PR #4452, opened 2026-08-29
Add asset directory to webpack watch folders

juliusstoerrle:
https://github.com/hitobito/hitobito/pull/4366 optimised webpack to only watch javascript sources, however it also processes the stylesheets with are now excluded from reloading. To improve developer experience during design activities, this PR proposes to include the /app/assets directory too.


carlobeltrame:
app/assets is the output directory for compiled assets, not an input directory. Editable CSS is in app/javascript/stylesheets which should lready be included in app/javascript..?

carlobeltrame:
Ah, nevermind, in the wagons it's in app/assets instead

juliusstoerrle:
If a better solution is to move wagon styles or use a narrower path, I can adjust the PR.

